PR #47, opened 2017-04-01
Added code to get the latest GO version

rjstyles:
Added code to automatically download the latest stable version of go without explicitly mentioning the version number.


kmova:
It would be good, if the code is as follows:

- If GO_VERSION is specified, download it. This might be required, when a released version patches are to be made without changing the dependencies. 
- If GO_VERSION line is commented (if empty or undeclared) the code should set the GO_VERSION as the latest GO_VERSION.

PR #52, opened 2017-04-13
added code to update the GO_VERSION if new version is available

rjstyles:


prateekpandey14:
Thanks @rjstyles ...Right

PR #63, opened 2017-07-17
Added few parameters

utkarshmani1997:
Added few parameters which are:
SectorSize, UsedLogicalBlocks and UsedBlocks
Now output have two tables:
1. Performance Stats
2. Capacity Stats

### DIFF
--- a/command/vsm_stats.go
+++ b/command/vsm_stats.go
@@ -406,6 +406,6 @@ func GetVolumeStats(address string, obj interface{}) (error, int) {
 		return err, -1
 	}
 	defer resp.Body.Close()
-	rc, _ := json.NewDecoder(resp.Body).Decode(obj), 0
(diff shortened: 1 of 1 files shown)


utkarshmani1997:
Updated the changes

command/vsm_stats.go
```diff
@@ -378,6 +406,6 @@ func GetVolumeStats(address string, obj interface{}) (error, int) {
 		return err, -1
 	}
 	defer resp.Body.Close()
-
-	return json.NewDecoder(resp.Body).Decode(obj), 0
+	rc, _ := json.NewDecoder(resp.Body).Decode(obj), 0
```
prateekpandey14: just decode it to the variable rc and return it with return statement with int

prateekpandey14:
LGTM... Thanks @utkarshmani1997

PR #81, opened 2017-09-07
Create ISSUE_TEMPLATE.md

prateekpandey14:
This commit will add issue template,project contributors will automatically see the template's contents in the issue form body. Templates customize and standardize the information we'd like included when contributors open issues.


docs/issue_template.md
```diff
@@ -0,0 +1,38 @@
...
+
+**What you expected to happen**:
+
+**How to reproduce it (as minimally and precisely as possible)**:
+
+
+**Anything else we need to know?**:
+
+**Environment**:
+- Maya version (use `maya version`):
```
kmova: Replace the maya related command with kubectl commands like for the template used in openebs/openebs

prateekpandey14: I think, these will be more specific to maya (and other maya-component), so it will be good to know that what version of maya and component they are running(in case - if they are running some different version altogether that will help us to debug).

PR #83, opened 2017-09-11
Fixing #296 and add error handling

utkarshmani1997:
* maya volume stats fails with one replica running.
* Error handling was not there for all the cases.

* from now on volume stats will query replica status also . Earlier this case was not there in `stats.go`.
* err does not throw proper error on `resp.StatusCode=500` i.e, it throws `nil` for this case. Now it will throw proper error.

* Run the new binary with one replica down/pending status and  output will be something like this
```IQN     : iqn.2016-09.com.openebs.jiva:vol1
Volume  : vol1
Portal  : 10.109.27.40:3260
Size    : 1G

     Replica|    Status|   DataUpdateIndex|
            |          |                  |
     Unknown|   Unknown|           Unknown|
   10.44.0.4|    Online|                 0|

----------- Performance Stats -----------

   r/s|   w/s|   r(MB/s)|   w(MB/s)|   rLat(ms)|   wLat(ms)|
     0|     0|     0.000|     0.000|      0.000|      0.000|

------------ Capacity Stats -------------

   Logical(GB)|   Used(GB)|
      0.000000|   0.000000|
```
* Set `resp.StatusCode` to `500` manually in stats.go and run the binary. you should get the proper error like "Server Unavailable" or "Internal Server Error".

### DIFF
--- a/command/volume_stats.go
+++ b/command/volume_stats.go
@@ -153,6 +153,7 @@ func (c *VsmStatsCommand) Run(args []string) int {
 	annotations, err := GetVolAnnotations(args[0])
 
 	if err != nil || annotations == nil {
+		fmt.Println(err)
 		return -1
 	}
 
@@ -283,8 +284,7 @@ func GetStatus(address string, obj interface{}) (error, int) {
 		return err, -1
 	}
 	defer resp.Body.Close()
-	rc := json.NewDecoder(resp.Body).Decode(obj)
-	return rc, 0
+	return json.NewDecoder(resp.Body).Decode(obj), 0
 }
 
 // NewControllerClient create the new replica client
--- a/command/volume_stats.go
+++ b/command/volume_stats.go
@@ -23,7 +23,7 @@ type Status struct {
 }
 
 type VolumeStats struct {
-	Cl              client.Resource
+	Resource        client.Resource
 	RevisionCounter int64         `json:"RevisionCounter"`
 	ReplicaCounter  int64         `json:"ReplicaCounter"`
 	SCSIIOCount     map[int]int64 `json:"SCSIIOCount"`
--- a/command/volume_stats.go
+++ b/command/volume_stats.go
@@ -137,7 +137,7 @@ func (c *VsmStatsCommand) Run(args []string) int {
 	)
 	statusArray := make([]string, 6)
 
-	flags := c.Meta.FlagSet("vsm-stats", FlagSetClient)
+	flags := c.Meta.FlagSet("volume stats", FlagSetClient)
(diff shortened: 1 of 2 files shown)


command/volume_stats.go
```diff
@@ -246,19 +272,19 @@ func GetStatus(address string, obj interface{}) (error, int) {
...
+		return errors.New("Server Not Reachable"), -1
 	}
 	if err != nil {
 		return err, -1
 	}
 	defer resp.Body.Close()
-
-	return json.NewDecoder(resp.Body).Decode(obj), 0
+	rc := json.NewDecoder(resp.Body).Decode(obj)
+	return rc, 0
```
prateekpandey14: Not required to store the decoded obj into rc, return it w/o rc

command/volume_stats.go
```diff
@@ -149,48 +151,72 @@ func (c *VsmStatsCommand) Run(args []string) int {
 	}
 
 	annotations, err := GetVolAnnotations(args[0])
+
 	if err != nil || annotations == nil {
 		return -1
```
prateekpandey14: Print the return error here

prateekpandey14: Thanks...

command/volume_stats.go
```diff
@@ -2,28 +2,28 @@ package command
...
 	"text/tabwriter"
+	"text/template"
 	"time"
 
 	"github.com/rancher/go-rancher/client"
 )
 
 type Status struct {
-	client.Resource
+	Cl              client.Resource
```
prateekpandey14: Changed this naming convention... ```Resource``` will be good

command/volume_stats.go
```diff
@@ -2,28 +2,28 @@ package command
...
 	ReplicaCounter  int64 `json:"replicacounter"`
 	RevisionCounter int64 `json:"revisioncounter"`
 }
 
 type VolumeStats struct {
-	client.Resource
+	Cl              client.Resource
```
prateekpandey14: Same here.

command/volume_stats.go
```diff
@@ -149,48 +151,73 @@ func (c *VsmStatsCommand) Run(args []string) int {
...
 	annotations, err := GetVolAnnotations(args[0])
+
 	if err != nil || annotations == nil {
+		fmt.Println(err)
 		return -1
 	}
+
 	if annotations.ControllerStatus != "Running" {
-		fmt.Println("Volume not reachable")
+		fmt.Printf("Volume not reachable")
```
prateekpandey14: Use Println() here.

prateekpandey14:
LGTM ..

PR #119, opened 2017-10-17
Use Alpine as base Docker image

ecliptik:

Reduces size of artifact Docker images by using [Alpine Linux](https://www.alpinelinux.org), comparisons between `ubuntu:16.04` and `alpine:3.6`

```
maya: 
ubuntu:16.04 
openebs/maya                 ci                  79b70f1da536        6 seconds ago       141MB 
 
alpine:3.6
openebs/maya                 ci                  e9a5dc4f58e6        7 seconds ago       24MB 
 
maya-apiserver: 
ubuntu:16.04 
openebs/m-apiserver          ci                  ca53afa02d2b        34 seconds ago      210MB 
 
alpine:3.6
openebs/m-apiserver          ci                  3f9a45ae621a        4 seconds ago       82.3MB
```

**Which issue this PR fixes** 

Fixes https://github.com/openebs/openebs/issues/287

Tested by running both images locally and applications started up properly using Alpine based container.


kmova:
@ecliptik -- that a significant improvement in size!

Couple of questions:
- Can we still retain installing "watch"?
- Can you provide a brief comment in there on why we are adding the new packages:
  - mii-tool
  - procps
  - libc6-compat

ecliptik:
@kmova 

There is no `watch` package in Alpine linux, so it will fail if it's added to the package list. `/bin/watch` is in the `procps` package for Alpine which is why it's added. In Ubuntu the `watch` package is a virtual package to `procps` as well,

https://packages.ubuntu.com/xenial/watch

The `net-tools` package in Alpine is comparable to the one in Ubuntu, however `mii-tool` is in it's own package.

Without `libc6-compat` the binary will give a "file not found" when trying to run on Alpine since some of the libraries built on a libc6 Linux system (like using when using the Vagrant dev VM) are not available by default on Alpine. Adding this package installs the required libraries for the binary to run on Alpine.

kmova:
Thanks @ecliptik 

Just one more question, earlier (am an year out-dated here), there used to be an "opinion" not to use alpine based images in production. Do you know if this view is changing now? Are there some examples of official docker images moving towards alpine..

kmova:
@ecliptik - A PR(https://github.com/openebs/maya/pull/116) got merged that resulted in the conflict, could you please update this PR by moving the file from buildscripts/docker/Dockerfile to buildscripts/mayactl/Dockerfile. 

Sorry about that :-(

ecliptik:
@kmova In regards to Alpine in production, we use it in a few different applications and it is stable as Ubuntu. The only major issue we've run into is dns resolution in musl libc when using service discovery within container orchestration. For the most part this has been resolved though (see https://github.com/gliderlabs/docker-alpine/issues/8).

Almost all [Docker Official Images](https://github.com/docker-library/official-images) have an Alpine variant. The only major difference right now is Alpine does not have full multi-arch support, which should be resolved shortly (https://github.com/gliderlabs/docker-alpine/issues/304).

ecliptik:
@kmova sync'd with master and Dockerfile is in the correct directory now

prateekpandey14:
@ecliptik , I still see some issues with DNS resolution in alpine (understood as much from the thread referred as well ). Here is my resolv.conf : 

```
nameserver 10.0.0.10
search default.svc.cluster.local svc.cluster.local cluster.local cbblr.com
options ndots:5
```
The issue renders even the CLI commands on the maya container way too slow (confirmed by setting the nameserver to 127.0.0.1, which immediately restores the speed - but we don't want this, right)

```
/ # date;maya;date
Thu Oct 19 15:23:40 UTC 2017
Usage: maya [--version] [--help] <command> [<args>]

Available commands are:
    snapshot    Provides operations related to snapshot of a Volume
    version     Prints version and other details relevant to maya
    volume      Provides operations related to a Volume

Thu Oct 19 15:24:02 UTC 2017
/ #
```

PR #125, opened 2017-10-23
Updated README.md with the required steps to build the binaries with Docker (Maya)

SiddharthMahapatra12:
In reference to issue #661 on openebs repo.


kmova:
@SiddharthMahapatra12  - Thanks for this initial content. However the approach for building with Docker is still under development. This PR can be re-visited once Docker builds are fully functional

PR #129, opened 2017-10-25
docker and build scripts for agent - #690

kmova:
This PR adds the initial version of the build and docker files for maya-agent.

Fixes https://github.com/openebs/openebs/issues/690

The build scripts are similar to maya-apiserver. Which also makes a case for merging all the build scripts together, which can be taken up in the future as part of : https://github.com/openebs/openebs/issues/499

### DIFF
--- a/buildscripts/agent/entrypoint.sh
+++ b/buildscripts/agent/entrypoint.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+set -ex
+
+MAYA_AGENT_NETWORK=$1
+
+CONTAINER_IP_ADDR=$(ip -4 addr show scope global dev "${MAYA_AGENT_NETWORK}" | grep inet | awk '{print $2}' | cut -d / -f 1)
+
+# Start agent service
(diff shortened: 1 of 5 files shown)


buildscripts/agent/entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+set -ex
+
+MAYA_AGENT_NETWORK=$1
+
+CONTAINER_IP_ADDR=$(ip -4 addr show scope global dev "${MAYA_AGENT_NETWORK}" | grep inet | awk '{print $2}' | cut -d / -f 1)
+
+# Start agent service
+exec /usr/local/bin/maya-agent up -bind="${CONTAINER_IP_ADDR}" 1>&2
```
utkarshmani1997: @kmova maya-agent up is not a command, it's maya-agent monitor.

utkarshmani1997: Why to bind when we can start it on localhost......and for communication purpose we will use k8s service

PR #169, opened 2017-10-31
Switch to glog for Volume API logging

prateekpandey14:
This commit will add glog as a default logger pkg for
volume API's for proper logging.

**Which issue this PR fixes** :
 fixes # openebs/openebs#265
ref - openebs/mayaserver#77

### DIFF
--- a/cmd/maya-apiserver/app/server/volume_endpoint.go
+++ b/cmd/maya-apiserver/app/server/volume_endpoint.go
@@ -15,22 +15,22 @@ import (
 //
 // TODO
 //    Should it return specific types than interface{} ?
-func (s *HTTPServer) VolumeSpecificRequest(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
+func (s *HTTPServer) volumeSpecificRequest(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
 
 	fmt.Println("[DEBUG] Processing", req.Method, "request")
 
 	switch req.Method {
 	case "PUT", "POST":
-		return s.VolumeAdd(resp, req)
+		return s.volumeAdd(resp, req)
 	case "GET":
-		return s.VolumeSpecificGetRequest(resp, req)
+		return s.volumeSpecificGetRequest(resp, req)
 	default:
 		return nil, CodedError(405, ErrInvalidMethod)
 	}
 }
 
 // VolumeSpecificGetRequest deals with HTTP GET request w.r.t a single Volume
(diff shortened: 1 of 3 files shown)


cmd/maya-apiserver/app/server/volume_endpoint.go
```diff
@@ -5,31 +5,32 @@ import (
...
+		return s.VolumeSpecificGetRequest(resp, req)
 	default:
 		return nil, CodedError(405, ErrInvalidMethod)
 	}
 }
 
-// vsmSpecificGetRequest deals with HTTP GET request w.r.t a single VSM
-func (s *HTTPServer) vsmSpecificGetRequest(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
+// VolumeSpecificGetRequest deals with HTTP GET request w.r.t a single Volume
+func (s *HTTPServer) VolumeSpecificGetRequest(resp http.ResponseWriter, req *http.Request) (interface{}, error) {
```
kmova: should this and the below functions be -- volumeSpecificGetRequest ?

cmd/maya-apiserver/app/server/volume_endpoint.go
```diff
@@ -87,23 +88,23 @@ func (s *HTTPServer) vsmList(resp http.ResponseWriter, req *http.Request) (inter
...
-func (s *HTTPServer) vsmRead(resp http.ResponseWriter, req *http.Request, vsmName string) (*v1.Volume, error) {
+// VolumeRead is the http handler that fetches the details of a Volume
+func (s *HTTPServer) VolumeRead(resp http.ResponseWriter, req *http.Request, volName string) (*v1.Volume, error) {
 
-	fmt.Println("[DEBUG] Processing Volume read request")
+	glog.Infof("Processing Volume read request")
 
-	if vsmName == "" {
+	if volName == "" {
 		return nil, CodedError(400, fmt.Sprintf("Volume name is missing"))
```
satyamz: I think you should log this error, Do you? There are other places like this!

cmd/maya-apiserver/app/server/volume_endpoint.go
prateekpandey14: Yes it should be.

cmd/maya-apiserver/app/server/volume_endpoint.go
prateekpandey14: `CodedError` accept only string right now as a arg, and glog don't have any error method which return string.

kmova:
@AmitKumarDas  -- should we cover the logging as part of this issue : https://github.com/openebs/openebs/issues/818

PR #182, opened 2017-11-08
Refactor Jiva controller and replica clients

prateekpandey14:
This PR will refactor the jiva controller and replica client codes 
and move inside to `pkg/client/jiva`.
This will help us to write CTL commands related to Volume and snapshots.

### DIFF
--- a/pkg/client/jiva/controller_client.go
+++ b/pkg/client/jiva/controller_client.go
@@ -1,0 +1,122 @@
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+)
+
+// NewControllerClient create the new controller client
+func NewControllerClient(address string) (*ControllerClient, error) {
+	address = strings.TrimPrefix(address, "tcp://")
+
+	if !strings.HasPrefix(address, "http") {
+		address = "http://" + address
+	}
+
+	if !strings.HasSuffix(address, "/v1") {
+		address += "/v1"
+	}
+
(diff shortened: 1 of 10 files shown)


pkg/client/jiva/model.go
```diff
@@ -50,7 +49,7 @@ type InfoReplica struct {
 	Chain           []string            `json:"chain"`
 	Disks           map[string]DiskInfo `json:"disks"`
 	RemainSnapshots int                 `json:"remainsnapshots"`
-	RevisionCounter string              `json:"revisioncounter"`
+	RevisionCounter int64               `json:"revisioncounter"`
```
kmova: why is this data type change required?

pkg/client/jiva/controller_client.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	if len(volume.Data) == 0 {
+		return nil, errors.New("No volume found")
+	}
+
+	return &volume.Data[0], nil
+}
+func (c *ControllerClient) Get(path string, obj interface{}) error {
+	//	resp, err := http.Get(c.address + path)
```
utkarshmani1997: remove commented code

pkg/client/jiva/model.go
prateekpandey14: It should be `string`, i think while refactoring/replace(duplicate struct), had deleted the struct which was getting used before.

prateekpandey14: done !

pkg/client/jiva/controller_client.go
prateekpandey14: done !!

PR #245, opened 2018-01-08
Move older volume stats cli to cobra cli

utkarshmani1997:
- Refers to issue openebs/openebs#505

- All the commands and flags now moved to maya/cmd/mayactl/command and are using
  the cobra library for the cli.

- Run mayactl inside maya-apiserver container to verify this change.
- Alternatively you can copy the binary inside the maya-apiserver and
  run the following commands
  * `mayactl volume stats -n=vol` to get in std I/O
  * `mayactl volume stats -n=vol -j=json` to get in json.

- none

5. Other details
improvement: part of #179

### DIFF
--- a/cmd/mayactl/app/command/stats_helper.go
+++ b/cmd/mayactl/app/command/stats_helper.go
@@ -1,0 +1,159 @@
+package command
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"time"
+)
+
+// Volume is a command implementation struct
+type Volume struct {
+	Spec struct {
+		AccessModes interface{} `json:"AccessModes"`
+		Capacity    interface{} `json:"Capacity"`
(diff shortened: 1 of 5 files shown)


cmd/mayactl/app/command/stats_helper.go
```diff
@@ -0,0 +1,159 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"time"
+)
+
+// Volume is a command implementation struct
+type Volume struct {
```
prateekpandey14: @utkarshmani1997  Forget to inform you , please use the `Volume{}` struct from the `types.go` files.

utkarshmani1997: @prateekpandey14 signature of both structures are different and there is some cleanup we have to do later regarding this.

prateekpandey14: ok

PR #254, opened 2018-01-31
Add unit test to improve coverage in pkg/client/mapiserver dir

utkarshmani1997:
- To improve test coverage
- It will reduce run time errors

- Adds unit tests in `pkg/client/mapiserver` dir

- Run `go test -v -cover`

- Few errors and unneccessary statements are removed and using the new errors
  created in `pkg/util/utils.go`.

5. Other details
improvement: openebs/openebs#1183

<!--

### DIFF
--- a/pkg/client/mapiserver/snapshot.go
+++ b/pkg/client/mapiserver/snapshot.go
@@ -39,7 +39,6 @@ func CreateSnapshot(volName string, snapName string) error {
 
 	_, err := GetStatus()
 	if err != nil {
-		err := fmt.Errorf("Unable to contact maya-apiserver: %s", GetURL())
 		return err
 	}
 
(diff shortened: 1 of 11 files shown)


pkg/client/mapiserver/snapshot.go
```diff
@@ -70,18 +69,9 @@ func CreateSnapshot(volName string, snapName string) error {
 	}
 	defer resp.Body.Close()
 
-	body, err := ioutil.ReadAll(resp.Body)
```
prateekpandey14: We have to read this body here , to get the proper error  which is return as a body

prateekpandey14: We have to read this body here , to get the proper error  which is return as a body

pkg/client/mapiserver/snapshot.go
```diff
@@ -70,18 +69,9 @@ func CreateSnapshot(volName string, snapName string) error {
...
-	if err != nil {
-		return err
-	}
-
 	code := resp.StatusCode
-	if err == nil && code != http.StatusOK {
```
prateekpandey14: This check required here in case of err is nil but there is err return in a body

pkg/client/mapiserver/snapshot.go
utkarshmani1997: Done

pkg/client/mapiserver/snapshot.go
utkarshmani1997: Done

PR #270, opened 2018-02-28
Get maya-api-service name at runtime

prateekpandey14:
This PR will get the maya-api-service name by using jsonpath and kubectl


kmova:
A good generic fix @prateekpandey14 

But I would like this step to also verify that maya-service is getting created with the namespace and release name prefixes.

PR #291, opened 2018-03-29
Fix mayactl volume stats command

utkarshmani1997:
`mayactl volume stats --volname <vol>` shows incorrect information

- To fix the issue openebs/openebs#1397
- Improve Unit test

- This commit fixes the issue by removing the unneccessary code.

- Create the volume with minimum three replicas
- Run `mayactl volume stats --volname <vol>` and it should show the
  desired output.

- none

5. Other details
- none

Signed-off-by: utkarshmani1997 <utkarshmani1997@gmail.com>

- fix: openebs/openebs#1397

<!--


avishnu:
Since the replica status block has been removed, what is the way to get the replica status?

PR #322, opened 2018-05-16
Add flag to specify the storage engine in maya-exporter

utkarshmani1997:
This commit adds the flag `storage.engine` or `-e` which can be
used to specify the storage engine while running the binary.

On branch feature-#1466
Changes to be committed:
	modified:   app/collector/collector.go
	modified:   app/command/command.go
	new file:   app/command/command_test.go
	modified:   app/command/exporter.go
	new file:   app/command/exporter_test.go
	modified:   main.go

- Refer to issue openebs/openebs#1312 and openebs/openebs#1466

- This commit adds the option for specifying the storage engine
  via a flag which can be passed while running the binary.
- This commit changes the type of size from int64 to float64.Hence
  now the output will not be zero if volume size is less than 1GB.

- Build and Run the binary with flag "-e" followed by storage engine
  and you should be able to get the result something like given below.
```
$ ./maya-exporter -e="cstor"
I0516 13:26:16.104162   26957 command.go:90] Starting maya-exporter ...
I0516 13:26:16.104235   26957 logs.go:43] maya-exporter does not support cstor yet
```
- For deployment purpose following changes are needed if storage engine
  is cstor else no change is required because jiva is the default storage
  engine.
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  name: maya-exporter
spec:
  replicas: 1
  template:
    metadata:
      labels:
        name: maya-exporter
    spec:
      # serviceAccountName: prometheus
      containers:
        - name: maya-exporter
          image: openebs/m-exporter:ci
          command: ["maya-exporter"]
          args:
            - "-e=cstor"
          ports:
          - containerPort: 9500
```

5. Other details
- Unit tests added which covers 42% of the statements.
- More effort needed to improve the code coverage which will be
  done in other PRs.

### DIFF
--- a/cmd/maya-exporter/app/collector/collector.go
+++ b/cmd/maya-exporter/app/collector/collector.go
@@ -313,9 +313,9 @@ func (e *VolumeExporter) collect() error {
 	aUsed = aUsed * sSize
 	aSize, _ := v1.DivideFloat64(aUsed, v1.BytesToGB)
 	actualUsed.Set(aSize)
-	size, _ := strconv.ParseInt(metrics1.Size, 10, 64)
-	size, _ = v1.DivideInt64(size, v1.BytesToGB)
-	sizeOfVolume.Set(float64(size))
+	size, _ := strconv.ParseFloat(metrics1.Size, 64)
+	size, _ = v1.DivideFloat64(size, v1.BytesToGB)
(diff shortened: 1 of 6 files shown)


utkarshmani1997:
This PR has to be raised on cstor-feature-integration, so closing this PR.

PR #326, opened 2018-05-16
Added unit test for getInClusterCS() in util_test.go

Subhamverma2407:
Signed-off-by: Subham Verma <subhamverma2407@gmail.com>
#### What this PR does / why we need it

* Add unit test function for getInClusterCS() in util_test.go
* To reduce run time logic errors.

####  Which issue this PR fixes 
This pr fixes issue #798 of openebs/openebs
https://github.com/openebs/openebs/issues/798


prateekpandey14:
Can you improve the test case, so that it follows the pattern explained in this wiki?https://github.com/golang/go/wiki/TableDrivenTests

IMO better to use the table approach (TDD), which will help you to cover both positive and negative code paths.

PR #345, opened 2018-05-29
Check the availability of the latest version of mayactl

### LINKED ISSUE #32: maya - option to check if a later version is available.
**_maya version_** command should also include a note to check if there is a later version available then the one that is currently installed. This will involve contacting the github releases page to get the latest version.

slalwani97:

- Checks the availability of the latest version of mayactl from the github releases page using git API. 
- Includes the unit test cases for the functions used.

**Which issue this PR fixes** : fixes #32 

screenshot from 2018-05-29 00-38-52

### DIFF
--- a/cmd/mayactl/app/command/version.go
+++ b/cmd/mayactl/app/command/version.go
@@ -19,9 +19,12 @@ package command
 import (
 	"fmt"
 	"runtime"
+	"strconv"
+	"strings"
 
 	"github.com/openebs/maya/orchprovider"
 	"github.com/openebs/maya/pkg/client/mapiserver"
+	"github.com/openebs/maya/pkg/util"
 	"github.com/openebs/maya/pkg/version"
 	"github.com/spf13/cobra"
 )
@@ -50,8 +53,67 @@ maya version
 			fmt.Println("m-apiserver status: ", mapiserver.GetConnectionStatus())
 
 			fmt.Println("Provider: ", orchprovider.DetectOrchProviderFromEnv())
+			util.CheckErr(checkLatestVersion(version.GetVersion()), util.Fatal)
 		},
 	}
 
 	return cmd
+}
+
+func checkLatestVersion(installedVersion string) error {
+	if installedVersion == "" {
+		return fmt.Errorf("GetVersion() returning empty string")
+	}
+
+	// removes first character i.e 'v' from the version
+	installedVersion = installedVersion[1:]
+
+	latestVersion, err := version.GetLatestRelease()
+	if err != nil {
+		return fmt.Errorf("found error - %s", err)
+	}
+
+	latest := parseVersion(latestVersion)
+	installed := parseVersion(installedVersion)
+
+	if latest == nil || installed == nil {
+		return fmt.Errorf("parseVersion() returning empty string")
+	}
+
+	flag := false
+
+	if installed[0] < latest[0] {
+		flag = true
+	} else if installed[1] < latest[1] && installed[0] == latest[0] {
+		flag = true
+	} else if installed[2] < latest[2] && installed[1] == latest[1] && installed[0] == latest[0] {
+		flag = true
+	}
+
+	if flag == true {
+		fmt.Println()
+		fmt.Println("A newer version of mayactl is available!")
+		fmt.Printf("Installed Version: v%s\n", installedVersion)
+		fmt.Printf("Latest version: v%s\n", latestVersion)
+	} else {
+		fmt.Println()
+		fmt.Println("You're running an up-to-date version of mayactl!")
+	}
+
+	return nil
+}
+
+func parseVersion(version string) []int64 {
+	versionList := strings.Split(version, ".")
+	versionNumber := []int64{}
+
+	for _, v := range versionList {
+		j, err := strconv.ParseInt(v, 10, 64)
+		if err != nil {
+			fmt.Printf("Unable to parse string to integer. \n Error - %s\n", err)
+			return nil
+		}
+		versionNumber = append(versionNumber, j)
+	}
+	return versionNumber
 }
--- a/cmd/mayactl/app/command/version.go
+++ b/cmd/mayactl/app/command/version.go
@@ -41,7 +41,6 @@ maya version
 	`,
 
 		Run: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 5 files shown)


cmd/mayactl/app/command/version.go
```diff
@@ -50,8 +53,67 @@ maya version
...
+
+	if flag == true {
+		fmt.Println()
+		fmt.Println("A newer version of mayactl is available!")
+		fmt.Printf("Installed Version: v%s\n", installedVersion)
+		fmt.Printf("Latest version: v%s\n", latestVersion)
+	} else {
+		fmt.Println()
+		fmt.Println("You're running an up-to-date version of mayactl!")
+	}
```
utkarshmani1997: i think "else" is not required, we have to print the details if the new version is available.
@prateekpandey14 @kmova what do you think ?

cmd/mayactl/app/command/version.go
```diff
@@ -50,8 +53,67 @@ maya version
...
+
+func parseVersion(version string) []int64 {
+	versionList := strings.Split(version, ".")
+	versionNumber := []int64{}
+
+	for _, v := range versionList {
+		j, err := strconv.ParseInt(v, 10, 64)
+		if err != nil {
+			fmt.Printf("Unable to parse string to integer. \n Error - %s\n", err)
+			return nil
```
utkarshmani1997: i think the std err is sufficient, need not to write the error message, just print the error.

cmd/mayactl/app/command/version.go
```diff
@@ -50,8 +53,67 @@ maya version
...
+	latestVersion, err := version.GetLatestRelease()
+	if err != nil {
+		return fmt.Errorf("found error - %s", err)
+	}
+
+	latest := parseVersion(latestVersion)
+	installed := parseVersion(installedVersion)
+
+	if latest == nil || installed == nil {
+		return fmt.Errorf("parseVersion() returning empty string")
```
utkarshmani1997: can the error msg be changed to "error in parsing string"

cmd/mayactl/app/command/version_test.go
```diff
@@ -0,0 +1,46 @@
...
+	utiltesting "k8s.io/client-go/util/testing"
+)
+
+var (
+	response2 = `[{"tag_name": "0.5.4"}]`
+)
+
+func TestCheckLatestVersion(t *testing.T) {
+	cases := []struct {
+		installed string
```
utkarshmani1997: can you use map[string]struct instead of []struct, the benefit of using map is you can write your test case name and it will increase the code readability too.

cmd/mayactl/app/command/version_test.go
```diff
@@ -0,0 +1,46 @@
...
+		{"0.5.7", "neg"},
+	}
+
+	fakeHandler :=
+		utiltesting.FakeHandler{
+			ResponseBody: string(response2),
+		}
+
+	for i, c := range cases {
+		server := httptest.NewServer(&fakeHandler)
```
utkarshmani1997: can you make use use of subtests instead of for loop. There are benefits of using the subtests.

pkg/version/version.go
```diff
@@ -78,3 +96,49 @@ func GetGitCommit() string {
 	}
 	return strings.TrimSpace(string(output))
 }
+
+// GetLatestRelease returns the latest version available
+func GetLatestRelease() (string, error) {
+
+	url := GitAPI + "/repos/openebs/maya/releases"
+
```
utkarshmani1997: use this as constant and move this into types/v1/constant.go

slalwani97:
Made the changes as requested.

cmd/mayactl/app/command/version.go
```diff
@@ -38,6 +41,7 @@ maya version
 	`,
 
 		Run: func(cmd *cobra.Command, args []string) {
+			util.CheckErr(checkLatestVersion(version.GetVersion()), util.Fatal)
 			fmt.Printf("Version: %s\n",
```
utkarshmani1997: let's move this to at the end of the execution hierarchy or run as go routine. @prateekpandey14 @AmitKumarDas @kmova what's your view on this ?

pkg/version/version.go
```diff
@@ -78,3 +98,49 @@ func GetGitCommit() string {
...
+		fmt.Printf("Unable to make a GET request. \n Error - %s\n", err)
+		return "", err
+	}
+
+	client := &http.Client{
+		Timeout: httpTimeout,
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		fmt.Printf("Request timeout occured. \n Error - %s\n", err)
```
prateekpandey14: Println() ?

cmd/mayactl/app/command/version.go
slalwani97: @utkarshmani1997 I think its better to call at the last.

pkg/version/version.go
```diff
@@ -78,3 +98,49 @@ func GetGitCommit() string {
...
 }
+
+// GetLatestRelease returns the latest version available
+func GetLatestRelease() (string, error) {
+
+	url := GitAPIAddr + v1.GitAPI
+
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		fmt.Printf("Unable to make a GET request. \n Error - %s\n", err)
```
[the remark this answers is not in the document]
slalwani97: done.

pkg/version/version.go
slalwani97: done.

ashishranjan738:
Hey i think some tests are failinig have a look .
screenshot from 2018-06-02 03-23-53

ashishranjan738:
I think there is some logical error with the code as the parse function converts string of integers to int64 so any leading 0's before a number will be removed . One corner testcase where it will fail is:-
LatestVersion: 0.5.2
CurrentVersion: 0.5.09
so after parsing the output for latest version will be []{0,5,2} and currentversion will be []{0,5,9} and after comparing the numbers of the int array the currentVersion will be greater than the latestVersion whereas the case is quite opposite .

sonasingh46:
@slalwani97 
Closing this PR due to inactivity. Please let us know if we need it.

PR #362, opened 2018-06-06
Improve volume stats using template and maintain unit test cases

slalwani97:

* Add templates for each table details.
* Uses template along with tabwriter for the proper alignment.
* Edit unit test cases accordingly.

**Which issue this PR fixes**: <https://github.com/openebs/openebs/issues/1597>

screenshot from 2018-06-06 12-45-45


slalwani97:
@kmova Sir, I found it better to give the header names in template only rather than hard coding the values in the map and shifting the index.

PR #368, opened 2018-06-11
Add pvc label in controller and replica deployment

Meghna-singh:

This PR adds a pvc label in the deployment specification files (controller and replica deployment).
Added pvc label in  createControllerDeployment() function in orchprovider/k8s.go 
https://docs.google.com/document/d/1voevyvyvEDfOW1IWjokKe9GVAGh0tmGCWZJSQH7QrBE/edit?usp=sharing

### DIFF
--- a/orchprovider/k8s/v1/k8s.go
+++ b/orchprovider/k8s/v1/k8s.go
@@ -800,6 +800,8 @@ func (k *k8sOrchestrator) createControllerDeployment(volProProfile volProfile.Vo
 		return nil, err
 	}
 
+	pvc := vol.Labels.K8sPersistentVolumeClaim
+
 	if clusterIP == "" {
 		return nil, fmt.Errorf("Volume cluster IP is required to create controller for volume 'name: %s'", vsm)
 	}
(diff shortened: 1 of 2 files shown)


orchprovider/k8s/v1/k8s.go
```diff
@@ -971,6 +975,8 @@ func (k *k8sOrchestrator) createReplicaDeployment(volProProfile volProfile.Volum
 		return nil, err
 	}
 
+	pvc := vol.Labels.K8sPersistentVolumeClaim
```
avishnu: Why do we need this change for the replica deployment?

orchprovider/k8s/v1/k8s.go
```diff
@@ -800,6 +800,8 @@ func (k *k8sOrchestrator) createControllerDeployment(volProProfile volProfile.Vo
...
```
avishnu: Can we get the pvc name from the volProProfile structure?

orchprovider/k8s/v1/k8s.go
avishnu: Ok.

Meghna-singh:
--> Before adding the pvc label.
image

--> After volume created through a pvc and adding pvc label .
screenshot from 2018-06-13 13-56-12
-->After volume created through mayactl and adding pvc label.
screenshot from 2018-06-13 14-01-33

types/v1/labels.go
```diff
@@ -485,6 +485,9 @@ const (
 	// VSMSelectorKey is used to filter vsm
 	VSMSelectorKey GenericAnnotations = "vsm"
 
+	//PVCSelectorKey is used to get pvc label
+	PVCSelectorKey GenericAnnotations = "pvc"
```
AmitKumarDas: This should be set as `openebs.io/pvc`

orchprovider/k8s/v1/k8s.go
Meghna-singh: we are using volume struct which is returned by the volume() method of volProProfile interface.
This volume struct holds the user's request for a OpenEBS volume.

PR #379, opened 2018-06-27
[cstor-integration] US2112 Fix IOPS, throughput calculation logic

utkarshmani1997:
On branch feature-#1465-cstor-exporter
Changes to be committed:
        modified:   cmd/maya-exporter/app/collector/cstorcollector.go
        modified:   cmd/maya-exporter/app/collector/cstorcollector_test.go
        modified:   cmd/maya-exporter/app/collector/jivacollector.go
        modified:   cmd/maya-exporter/app/collector/jivacollector_test.go
        modified:   cmd/maya-exporter/app/collector/metrics.go
        modified:   cmd/maya-exporter/app/command/command.go
        modified:   types/v1/metrics.go

-  The current logic of calculating IOPS and throughput in volume exporter
   needs to be enhanced. With the current logic, we are calculating IOPs
   values for 1 sec AFTER receiving the request from the Prom agent. Even
   though the agent requests every scrape interval, the IOs that have
   happened between the 2 scrapes are not considered for the IOPs.

-  In this approach make the exporter dumb i.e. it send the IO metric instead
   of IOPS, let the client take care of forming the logic for calculating IOPS
   from the successive IO values.

-  Deploy the maya-exporter on the cluster as sidecar and you should be
   able to get the metrics

-  This commit demands lot of changes in the configuration such as
   * PromQL should be changed to calculate the IOPS, throughput, latency etc
   * Dashboard should be changed, a new dashboard can be made with the new rules
   * Prometheus-config should also be changed (only namings).

5. Other details
-  Unit tests added which covers 30% and 92.8% in command and collector
   package respectively.
-  Unified the design for jiva and cstor.


cmd/maya-exporter/app/collector/metrics.go
```diff
@@ -127,112 +137,211 @@ type MetricsDiff struct {
 // MetricsInitializer returns the Metrics instance used for registration
 // of exporter while instantiating JivaStatsExporter and
 // CstorStatsExporter.
-func MetricsInitializer() *Metrics {
+func MetricsInitializer(casType string) *Metrics {
 	return &Metrics{
```
avishnu: Re-order the metrics into:
1) Uptime
2) Capacity-relate metrics
3) Performance-related metrics grouped by reads and writes.

cmd/maya-exporter/app/collector/metrics.go
```diff
@@ -127,112 +137,211 @@ type MetricsDiff struct {
...
+// It tries to reconnect with the volume if there is any error via a goroutine.
+func (v *VolumeStatsExporter) Collect(ch chan<- prometheus.Metric) {
+	// no need to catch the error as exporter should work even if
+	// there are failures in collecting the metrics due to connection
+	// issues or anything else.
+	switch v.CASType {
+	case "cstor":
+		_ = v.Cstor.collector(v)
+	case "jiva":
+		_ = v.Jiva.collector(v)
```
avishnu: Maybe better to pass v.Metrics instead of the entire top-level object v.

cmd/maya-exporter/app/collector/metrics.go
```diff
@@ -127,112 +137,211 @@ type MetricsDiff struct {
...
-		}),
-
-		volumeUpTime: prometheus.NewGaugeVec(prometheus.GaugeOpts{
-			Namespace: "OpenEBS",
-			Name:      "volume_uptime",
-			Help:      "Time since volume has registered",
-		},
+		totalReadBytes: prometheus.NewGauge(
+			prometheus.GaugeOpts{
+				Namespace: "OpenEBS",
```
avishnu: Namespace should be set to the actual namespace where the volume controller and exporter are running. This can be addressed in a separate PR.

cmd/maya-exporter/app/collector/metrics.go
```diff
@@ -127,112 +137,211 @@ type MetricsDiff struct {
...
-			Name:      "size_of_volume",
-			Help:      "Size of the volume requested",
-		}),
-
-		volumeUpTime: prometheus.NewGaugeVec(prometheus.GaugeOpts{
-			Namespace: "OpenEBS",
-			Name:      "volume_uptime",
-			Help:      "Time since volume has registered",
-		},
+		totalReadBytes: prometheus.NewGauge(
```
avishnu: Re-order the metrics into:
1. Capacity-related metrics.
2. Performance-related metrics grouped by reads and writes.
3. Counter metrics.

cmd/maya-exporter/app/collector/jivacollector.go
```diff
@@ -6,122 +6,44 @@ package collector
...
-	// collect the metrics extracted by collect method
-	for _, gauge := range j.gaugesList() {
-		gauge.Collect(ch)
-	}
-	for _, counter := range j.countersList() {
-		counter.Collect(ch)
-	}
-}
-
 // getVolumeStats is used to decode the response from the Jiva controller
```
avishnu: Since we are now using unmarshal, we can replace decode with unmarshal.

cmd/maya-exporter/app/collector/cstorcollector.go
```diff
@@ -157,7 +66,7 @@ func (c *CstorStatsExporter) writer() error {
 
 // reader reads the response from socket in the buffer and
 // exits if the buffer contains the "IOSTATS\r\n" at the end.
-func (c *CstorStatsExporter) reader() (string, error) {
+func (c *Cstor) reader() (string, error) {
 	buf := make([]byte, 256)
```
avishnu: make 256 as a constant READ_BUF_SIZE

types/v1/metrics.go
```diff
@@ -24,6 +26,29 @@ type VolumeMetrics struct {
 	Name              string  `json:"Name"`
 }
 
+// VolumeStats is used to store the collected metrics
+// TODO: Make this generic, so that it can be used by mayactl
+// and other components of maya.
+type VolumeStats struct {
+	Iqn                 string      `json:"iqn"`
+	Reads               json.Number `json:"ReadIOPS"`
```
avishnu: json tag should also be changed to Reads instead of ReadIOPS.

types/v1/metrics.go
```diff
@@ -24,6 +26,29 @@ type VolumeMetrics struct {
...
+	TotalReadTime       json.Number `json:"TotalReadTime"`
+	TotalReadBlockCount json.Number `json:"TotalReadBlockCount"`
+	TotalReadBytes      json.Number `json:"TotalReadBytes"`
+
+	Writes               json.Number `json:"WriteIOPS"`
```
avishnu: json tag should also be changed to Writes instead of WriteIOPS.

cmd/maya-exporter/app/collector/metrics.go
utkarshmani1997: @avishnu There are two approaches to get the namespace by the application
1. [Use downward API's](https://github.com/kubernetes/kubernetes/blob/release-1.0/docs/user-guide/downward-api.md) (simple and easiest ones)
2. Make call to kube-apiserver to get the namespace by use of kubernetes client.

types/v1/metrics.go
utkarshmani1997: This demands the changes in Jiva, mayactl and istgt as well and hence things may break. So it will be better if we can create a separate task for this and address this in new PR.

types/v1/metrics.go
utkarshmani1997: same comments as above

PR #406, opened 2018-07-26
[TA2114] remove subsystem name from metrics queries

utkarshmani1997:
TA2114-remove-subsystem-name
    Changes to be committed:
        modified:   cmd/maya-exporter/app/collector/cstorcollector.go
        modified:   cmd/maya-exporter/app/collector/cstorcollector_test.go
        modified:   cmd/maya-exporter/app/collector/jivacollector.go
        modified:   cmd/maya-exporter/app/collector/jivacollector_test.go
        modified:   cmd/maya-exporter/app/collector/metrics.go

1. Why is this change necessary?
-  To make the query unique (same) for both jiva and cstor.

2. How does it address the issue?
-  This commit removes the subsystem name from the query and change the
       namespace suffix from `OpenEBS` to `openebs`

3. What side effects does this change have?
-  This commit will require the change in the PromQL for metrics collection.

4. How to verify this change?
-  you can query the metrics `openebs_size_of_volume` and you will be able to
   get the size of volume.You can also query the other metrics likewise.

5. Any specific message for reviewer ?
-  A PR will be raised in `openebs/openebs` which will address the changes
   required for this commit.


cmd/maya-exporter/app/collector/metrics.go
```diff
@@ -143,114 +143,100 @@ func MetricsInitializer(casType string) *Metrics {
...
 		volumeUpTime: prometheus.NewCounterVec(
 			prometheus.CounterOpts{
-				Namespace: "OpenEBS",
-				Subsystem: casType,
+				Namespace: "openebs",
 				Name:      "volume_uptime",
 				Help:      "Time since volume has registered",
 			},
-			[]string{"volName", "iqn", "portal"},
+			[]string{"volName", "iqn", "portal", "castype"},
```
avishnu: Should it be storageengine or castype?

avishnu: Is it better if it is renamed to storageenginetype rather than castype?

avishnu: Ok

PR #430, opened 2018-08-02
using environment variable for getting the default namespace to look for CRs

Moteesh:
Also cleaned up some of the log messages


Moteesh:
This feature is no longer required. So closing it.

PR #440, opened 2018-08-08
Handle pool deletion for all kubectl version

sonasingh46:
This PR will do following:

1. Capture delete event from deleteFunc hook.

2. Supress delete event from updateFunc hook
   that might happen for some kubectl version.

3. Add unit test for newer code.

4. Comments and update of variable name.


cmd/maya-apiserver/spc-watcher/controller.go
```diff
@@ -128,19 +145,31 @@ func (c *Controller) updateSpc(oldSpc, newSpc interface{}) {
 		// If Resource Version is same it means the object has not got updated.
 		c.queueLoad.Operation = ignoreEvent
 	} else {
+		// Suppressing delete event here as the event is already captured in
+		// deleteSpc hook.
 		if IsDeleteEvent(spcObjectNew) {
-			c.queueLoad.Operation = deleteEvent
+			c.queueLoad.Operation = ignoreEvent
```
avishnu: Please add a log that we are ignoring the delete event here as it will be handled as part of delete operation.

cmd/maya-apiserver/spc-watcher/handler.go
```diff
@@ -39,11 +39,11 @@ func (c *Controller) syncHandler(key, operation string, object interface{}) erro
 	// Check if the event is for delete and use the spc object that was pushed in the queue
 	// for utilising details from it e.g. delete cas template name for storagepool deletion.
 	if operation == deleteEvent {
-		object := object.(*apis.StoragePoolClaim)
-		if object == nil {
+		spcObject := object.(*apis.StoragePoolClaim)
+		if spcObject == nil {
```
avishnu: Please add a comment that object interface is never nil, hence the nil check is done after typecast only.

sonasingh46: Sure..

cmd/maya-apiserver/spc-watcher/controller.go
sonasingh46: Sure

cmd/maya-apiserver/spc-watcher/handler.go
sonasingh46: @avishnu Little bit of more information.(To correct myself)
The concept is following:
```Interface type of nil is different from nil but all other type of nil has the same type  as that of nil.```
That means, if you pass a nil object of type let us say, `A` to an interface and then try to compare the object for equality with nil,
it will return false.

PR #455, opened 2018-08-17
feat (snapshot-grpc) Server side gRPC implementation for snapshots updated with requested changes

moteesh:
cstor volume grpc server for snapshot operations
    changes since last review:
        - removed ssh and syslogd services from the volume-mgmt container
        - added some code level comments for IoWaitTime and TotalWaitTime
        - moved cstorvolume.pb.go to pkg/client/generated/cstor-volume-grpc package
        - returning error from the createsnapshot and destroysnapshot
        methods upon receiving an error in the snapshot command response
        - removed the unnecessary status in the snapshot command request
        - sending status of commands in json format in the response
        fixed an issue in --port flag parsing for server start command
        - making the grpc listen port configurable through command line option
        - avoid generating .pb.go files in travis and do it in dev environment only
        - removed net-tools package from docker image as we do not need it
        - usining consistent logging library i.e glog eveywhere
        - moved grpc proto file to v1alpha and renamed to cstorvolume.proto


cmd/cstor-volume-grpc/app/command/start.go
```diff
@@ -0,0 +1,65 @@
...
+//MaxPortNumber that can be used for ipv4
+const MaxPortNumber = 65535
+
+// CmdStartOptions has flags for starting CStorVolume watcher.
+type CmdStartOptions struct {
+	port string
+}
+
+// Validate validates the flag values
+func (c *CmdStartOptions) Validate(cmd *cobra.Command) error {
```
gila: Not sure why we do this -- net.Listen() would bail out if you pass it illegal stuff.

pkg/apis/openebs.io/v1alpha1/cstorvolume.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
```
gila: Okay, this looks better already.  As we now have actual messages defined for specifics requests, thanks

pkg/apis/openebs.io/v1alpha1/cstorvolume.proto
```diff
@@ -0,0 +1,17 @@
...
+
+package v1alpha1;
+
+message VolumeSnapRequest {
+	string command = 1;
```
gila: Dont think this is needed. Its an implementation detail for cstor

pkg/apis/openebs.io/v1alpha1/cstorvolume.proto
```diff
@@ -0,0 +1,17 @@
...
+	string volume = 2;
+	string snapname = 3;
+}
+
+message VolumeSnapResponse {
+	bytes status = 1;
+}
+
+service RunSnapCommand {
```
gila: We should create snapshot, clone, etc methods here right? Then also we dont need to command string ..

cmd/cstor-volume-grpc/api/handler.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+//APIUnixSockVar is unix socker variable
+var APIUnixSockVar util.UnixSock
+
+// Server represents the gRPC server
+type Server struct {
+}
+
+// RunVolumeSnapCommand generates response to a RunCommand request
```
gila: I'd expect methods for create, destroy, list etc. not only mega method that does a big switch/if right?

PR #463, opened 2018-08-22
Create ZFS volume with compression, thin provisioning and with 4k blocksize by default

Moteesh:
- added -s option for  enabling thin-provisioning
- using 4K as the default volblocksize instead of 32K
- enabling compression by default with option compression=on


cmd/cstor-pool-mgmt/volumereplica/volumereplica.go
```diff
@@ -76,6 +76,7 @@ func createVolumeBuilder(cStorVolumeReplica *apis.CStorVolumeReplica, fullVolNam
 	openebsTargetIP := "io.openebs:targetip=" + cStorVolumeReplica.Spec.TargetIP
 
 	createVolAttr = append(createVolAttr, "create",
+		"-b", "4K", "-s", "-o", "compression=on",
```
avishnu: Is it fine to pass the "-o compression=on" in the middle instead of at the end ?

vishnuitta: this is fine

PR #467, opened 2018-08-22
Use patch instead of update for cvr finalizer.

princerachit:
Subsequent delete calls causes the resource version
of the cstorvolumereplica to change.

When the first delete call is made the fetched cvr's
finalizer field is updated locally, meanwhile another
request arrives and the resource version of cvr in
k8s changes.

When the code tries to update cvr using the first
object it fetched the update fails due to rv mismatch.

To overcome this issue I have changed the update call
to patch. This would not be affected by multiple calls.

Changes:
modified:   cmd/cstor-pool-mgmt/controller/replica-controller/handler.go
modified:   cmd/cstor-pool-mgmt/controller/replica-controller/new_replica_controller.go
modified:   cmd/cstor-pool-mgmt/volumereplica/volumereplica.go

### DIFF
--- a/cmd/cstor-pool-mgmt/controller/replica-controller/handler.go
+++ b/cmd/cstor-pool-mgmt/controller/replica-controller/handler.go
@@ -17,6 +17,7 @@ limitations under the License.
 package replicacontroller
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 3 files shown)


cmd/cstor-pool-mgmt/controller/replica-controller/handler.go
```diff
@@ -98,6 +123,7 @@ func (c *CStorVolumeReplicaController) cVREventHandler(operation common.QueueOpe
 
 		err := volumereplica.DeleteVolume(fullVolName)
 		if err != nil {
+			glog.Errorf("Error in deleting volume %q: %s", err)
```
avishnu: Missing variable for %q format-specifier.

PR #472, opened 2018-08-24
Improved mayactl error message when volumeInfo UT

ashishranjan738:

-> Improved error message when volume not found at maya-apiserver
-> Added UT for volumes.go function .

### DIFF
--- a/cmd/mayactl/app/command/volume.go
+++ b/cmd/mayactl/app/command/volume.go
@@ -223,7 +223,15 @@ func (volInfo *VolumeInfo) GetControllerStatus() string {
 
 // GetIQN returns the IQN of the volume
 func (volInfo *VolumeInfo) GetIQN() string {
-	return volInfo.Volume.Spec.Iqn
+	if len(volInfo.Volume.Spec.Iqn) > 0 {
+		return volInfo.Volume.Spec.Iqn
+	} else if val, ok := volInfo.Volume.ObjectMeta.Annotations["openebs.io/iqn"]; ok {
+		return val
+	} else if val, ok := volInfo.Volume.ObjectMeta.Annotations["vsm.openebs.io/iqn"]; ok {
+		return val
+	}
+	return ""
+
 }
 
 // GetVolumeName returns the volume name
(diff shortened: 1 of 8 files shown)


cmd/mayactl/app/command/volume.go
```diff
@@ -69,13 +69,13 @@ func NewVolumeInfo(URL string, volname string, namespace string) (volInfo *Volum
 	}
 	if resp != nil && resp.StatusCode != 200 {
 		if resp.StatusCode == 500 {
-			fmt.Printf("Volume: %s not found at M_API server\n", volname)
+			fmt.Printf("Volume: %s not found at namespace: %q\n", volname, namespace)
 			err = util.InternalServerError
 		} else if resp.StatusCode == 503 {
 			fmt.Println("M_API server not reachable")
```
kmova: Can you fix this also to print what is the IP address used to connect. Rename --> "M_API server" to maya apiservice

cmd/mayactl/app/command/volume.go
```diff
@@ -69,13 +69,13 @@ func NewVolumeInfo(URL string, volname string, namespace string) (volInfo *Volum
...
 			err = util.ServerUnavailable
 		} else if resp.StatusCode == 404 {
-			fmt.Printf("Volume: %s not found at M_API server\n", volname)
+			fmt.Printf("Volume: %s not found at namespace: %q error: %s\n", volname, namespace, http.StatusText(resp.StatusCode))
 			err = util.PageNotFound
 		}
 		fmt.Printf("Received an error from M_API server: statuscode: %d", resp.StatusCode)
```
kmova: M_API server -- check for this occurrence and change to "maya apiserver"

cmd/mayactl/app/command/volume.go
ashishranjan738: sure sir

cmd/mayactl/app/command/volume.go
ashishranjan738: sure

cmd/mayactl/app/command/volume_test.go
```diff
@@ -49,32 +55,38 @@ func TestGetCASType(t *testing.T) {
...
-			Volume: CASVolume{
-				ObjectMeta: metav1.ObjectMeta{
-					Annotations: map[string]string{
-						"vsm.openebs.io/cluster-ips": "192.168.100.1",
+			Volume: VolumeInfo{
+				Volume: v1alpha1.CASVolume{
+					ObjectMeta: metav1.ObjectMeta{
+						Annotations: map[string]string{
+							"vsm.openebs.io/cluster-ips": "192.168.100.1",
+						},
```
sonasingh46: Add both `"openebs.io/cluster-ips": "192.168.100.1"` and `"vsm.openebs.io/cluster-ips": "192.168.100.1"` in VolumeInfo object so test which get s selected.

cmd/mayactl/app/command/volume_test.go
```diff
@@ -137,15 +155,17 @@ func TestGetControllerStatus(t *testing.T) {
...
-				},
-				Spec: CASVolumeSpec{
-					Iqn: "iqn.2016-09.com.openebs.cstor:default-testclaim7",
+			Volume: VolumeInfo{
+				Volume: v1alpha1.CASVolume{
+					ObjectMeta: metav1.ObjectMeta{
+						Annotations: map[string]string{},
+					},
+					Spec: v1alpha1.CASVolumeSpec{
+						Iqn: "iqn.2016-09.com.openebs.cstor:default-testclaim7",
```
sonasingh46: What if the spec is emtpy ? How do we handle it for 0.6 ?

prateekpandey14: In 0.6 there is no spec so its fine

sonasingh46: But then ww need to find the value from annotation?
Cant it be like this that if not found in spec fetch from annotations?

ashishranjan738: added to check for annotation if not present in spec

cmd/mayactl/app/command/volume_test.go
ashishranjan738: sure

cmd/mayactl/app/command/volume_test.go
sonasingh46: That is great

cmd/mayactl/app/command/volume_test.go
sonasingh46: Thanks

sonasingh46: The functions getClusterIP amd getControllerStatus does not try to fetch from specs but only
from anmotations.
Is it so that for every case these values will always be present in annotations?

ashishranjan738: No it will be present on annotation everytime

PR #474, opened 2018-08-24
Fix mayactl for cstor and jiva volumes

prateekpandey14:
This PR includes the changes that were initially developed and merged under #459. However due to the bug in getting the volume info, the commit has been reverted from the master. 

This PR will be used to track the fix for mayactl info and stats command on jiva volumes. 

This PR also fixes the CI scripts that need to handle change in PV name which has now a random generated string suffix. 


ci/install_openebs.sh
```diff
@@ -43,8 +43,9 @@ kubectl get pvc,pv
 
 kubectl get pods --all-namespaces
 
+JIVACTRL=$(kubectl get deploy -l openebs/controller=jiva-controller --no-headers | awk {'print $1'})
```
kmova: can you change the label to "openebs.io/controller=jiva-controller". Similarly with the replica.

PR #487, opened 2018-08-28
Refactor code and add unit tests

sonasingh46:
Refactor code and add unit tests. 

### DIFF
--- a/cmd/maya-apiserver/spc-watcher/select_disk.go
+++ b/cmd/maya-apiserver/spc-watcher/select_disk.go
@@ -24,7 +24,6 @@ import (
 	"github.com/golang/glog"
 	"github.com/openebs/maya/pkg/apis/openebs.io/v1alpha1"
 	openebs "github.com/openebs/maya/pkg/client/generated/clientset/internalclientset"
-	"github.com/openebs/maya/pkg/client/k8s"
 )
 
 // clientset struct holds the interface of internalclientset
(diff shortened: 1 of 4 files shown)


cmd/maya-apiserver/spc-watcher/select_disk.go
```diff
@@ -185,12 +154,12 @@ func (k *clientSet) nodeSelector(listDisk *v1alpha1.DiskList, poolType string, s
 			// Entry to this block means the hostname was already mapped and it has more than one disk and at least two disks.
 			nodeDisk := nodeDiskMap[value.Labels[string(v1alpha1.HostNameCPK)]]
 			// Increment the disk count
-			nodeDisk.diskCount++
+			//nodeDisk.diskCount++
```
avishnu: This comment and the above comment may not be needed, can be removed.

cmd/maya-apiserver/spc-watcher/storagepool_create.go
```diff
@@ -54,8 +55,24 @@ func CreateStoragePool(spcGot *apis.StoragePoolClaim, reSync bool, pendingPoolCo
 		return nil
 	}
 
+	// Get kubernetes clientset
```
avishnu: Maybe we should initialise the clientset in spcSync(), which is the caller function of createStoragePool and bind createStoragePool to the structure. As you suggested @sonasingh46 , it may also add value to rename the structure to a more apt name.

sonasingh46: Will rethink on this part... Need to discuss its packaging, code structure, naming conventions etc on a bigger picture with team. But will add a backlog for this.

avishnu: Ok

PR #489, opened 2018-08-28
Fix error message when volume no found in namespaces

ashishranjan738:

-> This fix error message when volume is not found in namespace.

### DIFF
--- a/cmd/mayactl/app/command/volume.go
+++ b/cmd/mayactl/app/command/volume.go
@@ -163,12 +163,15 @@ func NewVolumeInfo(URL string, volname string, namespace string) (volInfo *Volum
 		if resp.StatusCode == 500 {
 			fmt.Printf("Volume: %s not found at namespace: %q\n", volname, namespace)
 			err = util.InternalServerError
+			return
 		} else if resp.StatusCode == 503 {
 			fmt.Printf("maya apiservice not reachable at %q\n", mapiserver.GetURL())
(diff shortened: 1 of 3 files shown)


cmd/mayactl/app/command/volume_info.go
```diff
@@ -102,7 +102,7 @@ func (c *CmdVolumeOptions) RunVolumeInfo(cmd *cobra.Command) error {
 	// controller's IP, status, iqn, replica IPs etc.
 	volumeInfo, err := NewVolumeInfo(mapiserver.GetURL()+VolumeAPIPath+c.volName, c.volName, c.namespace)
 	if err != nil {
-		return err
+		return nil
```
ashishranjan738: returning nil because we want to mayactl to exit with 0 status code.

cmd/mayactl/app/command/volume.go
```diff
@@ -163,12 +163,15 @@ func NewVolumeInfo(URL string, volname string, namespace string) (volInfo *Volum
 		if resp.StatusCode == 500 {
 			fmt.Printf("Volume: %s not found at namespace: %q\n", volname, namespace)
```
kmova: Can you check with @mahebbar how to work this error. Should be different from 404.

ashishranjan738: Liked ur message so updated that

PR #496, opened 2018-08-28
Update Jiva and cStor CAS templates for 0.7.0

kmova:
In this PR, the default CAST YAMLs, StoragePool and StorageClasses
installed by maya-apiserver are updated with the recent
changes done to the openebs-pre-release.yaml

As part of moving from the openebs/openebs repo, the
following changes have been made:
- use the ENV to fill up some of the default values
  like namespace, service account name, image ids, etc.,
- make the content parsable for strings from go-code, by
  making changes like:
  * Replcace back-quote ` with double quotes in jsonpath
  * Escape the . and " within jsonpath
  * replace string false with none
  * quote the replica count coming from env to string
- removed the older/deprecated CAST YAMLs from install folder

Also, updated the travis ci scripts to skip installing the
pre-release.yaml. After the 0.7.0 release is done, this
file can be re-enabled back


pkg/install/v1alpha1/cstor_pool_0.7.0.go
```diff
@@ -386,25 +461,25 @@ metadata:
...
+    id: liststoragepoolcr
     apiVersion: openebs.io/v1alpha1
     kind: StoragePool
     action: list
     options: |-
       labelSelector: openebs.io/storage-pool-claim={{.Storagepool.owner}}
   post: |
-    {{- $sps := jsonpath .JsonResult "{range .items[*]}pkey=sps,{@.metadata.name}=;{end}" | trim | default "" | splitList ";" -}}
-    {{- $sps | notFoundErr "storge pool cr not found" | saveIf "deletelistcsp.notFoundErr" .TaskResult | noop -}}
+    {{- $sps := jsonpath .JsonResult "{range .items[*]}pkey=sps,{@.metadata.name}="";{end}" | trim | default "" | splitList ";" -}}
```
sonasingh46: This will fail, I guess.
`=""`
Make it to `=` only.
The intention of `""` is to push empty. If you dont put `""` it will still put empty

sonasingh46: Please check for other such cases..

The only changes required to pull pre release yaml here are following to make runtasks work
1. Replace ` ` with " "
2. Escape ```. ``` charcters (e.g. in label keys ) by putting \\ befor the ```. ```character
3. The one specified above.

pkg/install/v1alpha1/cstor_sparse_pool_claim_0.7.0.go
```diff
@@ -48,10 +48,10 @@ apiVersion: storage.k8s.io/v1
 kind: StorageClass
 metadata:
   name: openebs-cstor-sparse
+  labels:
+    openebs.io/cas-type: cstor
   annotations:
-    cas.openebs.io/create-volume-template: cstor-volume-create-default-0.7.0
-    cas.openebs.io/delete-volume-template: cstor-volume-delete-default-0.7.0
-    cas.openebs.io/read-volume-template: cstor-volume-read-default-0.7.0
+    openebs.io/cas-type: cstor
```
avishnu: Why do we need to have both label and annotation ?

kmova: make sense. removed from labels. annotation is the one used by the code to map type to templates.

pkg/install/v1alpha1/cstor_pool_0.7.0.go
kmova: Thank you! addressed.

ci/install_openebs.sh
```diff
@@ -18,12 +22,19 @@ for i in $(seq 1 50) ; do
...
-MAPIPOD=$(kubectl get pods -o jsonpath='{.items[?(@.spec.containers[0].name=="maya-apiserver")].metadata.name}' -n openebs)
-
 echo "--------------- Maya apiserver initial logs -----------------------------"
+MAPIPOD=$(kubectl get pods -o jsonpath='{.items[?(@.spec.containers[0].name=="maya-apiserver")].metadata.name}' -n openebs)
 kubectl logs $MAPIPOD -n openebs
 printf "\n\n"
 
+#Print the default StoragePools Created
+kubectl get sp
+sleep 1
```
princerachit: Why sleep?

pkg/install/v1alpha1/cstor_volume_0.7.0.go
```diff
@@ -233,27 +277,40 @@ spec:
...
+          app: cstor-volume-manager
           openebs.io/target: cstor-target
           openebs.io/persistent-volume: {{ .Volume.owner }}
-          app: cstor-volume-manager
       template:
         metadata:
           labels:
             {{- if eq $isMonitor "true" }}
             monitoring: volume_exporter_prometheus
+            openebs_pv: {{ .Volume.owner }}
```
princerachit: Are we following `_` convention?

kmova: Added for backward compability with prometheus config. When it gets changed to just use openebs.io/persistent-volume, we can remove this.

ci/install_openebs.sh
kmova: good catch. removed.

PR #499, opened 2018-08-30
 Extend error message and  print response error

prateekpandey14:
This commit will enhance the error logging , will read the response from http response body and print in case of http response status is not OK(200) 

### DIFF
--- a/pkg/client/mapiserver/volumes_list.go
+++ b/pkg/client/mapiserver/volumes_list.go
@@ -13,7 +13,7 @@ const (
 // ListVolumes and return them as obj
 func ListVolumes(obj interface{}) error {
 
-	body, err := getRequest(GetURL()+listVolumePath, "", false)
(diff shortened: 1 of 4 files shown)


pkg/client/mapiserver/volumes_list.go
```diff
@@ -13,7 +13,7 @@ const (
 // ListVolumes and return them as obj
 func ListVolumes(obj interface{}) error {
 
-	body, err := getRequest(GetURL()+listVolumePath, "", false)
+	body, err := getRequest(GetURL()+listVolumePath, "", true)
```
kmova: what does `true` or `false` here mean?

prateekpandey14: Its for getting the response from the http response body in case of any error...basically non 200 response

prateekpandey14: if true it will read the response from body ...

PR #522, opened 2018-09-19
feat (snapshot-grpc) server side grpc implementation for cstor volume snapshots with all review comments incorporated.

Moteesh:
- need to close PRs https://github.com/openebs/maya/pull/455 and https://github.com/openebs/maya/pull/439.
- all the review comments given in the above two PRs have been addressed in this PR.


cmd/cstor-volume-grpc/api/handler_test.go
```diff
@@ -0,0 +1,95 @@
...
+		"successSnapshotCreate": {
+			expectedError: nil,
+			test: &v1alpha1.VolumeSnapCreateRequest{
+				Version:  ProtocolVersion,
+				Volume:   "dummyvol1",
+				Snapname: "dummysnap1",
+			},
+		},
+	}
+
```
prateekpandey14: Can we add failure or negative test cases, will be more helpful.

cmd/cstor-volume-grpc/api/handler_test.go
```diff
@@ -0,0 +1,95 @@
...
+		"successSnapshotDestroy": {
+			expectedError: nil,
+			test: &v1alpha1.VolumeSnapDeleteRequest{
+				Version:  ProtocolVersion,
+				Volume:   "dummyvol1",
+				Snapname: "dummysnap1",
+			},
+		},
+	}
+
```
prateekpandey14: Same as above comment.

cmd/cstor-volume-grpc/api/handler.go
```diff
@@ -0,0 +1,90 @@
...
+	return resp, err
+}
+
+// DeleteSnapshot sends snapdelete command to istgt and returns the response
+func DeleteSnapshot(ctx context.Context, in *v1alpha1.VolumeSnapDeleteRequest) (*v1alpha1.VolumeSnapDeleteResponse, error) {
+	sockresp, err := APIUnixSockVar.SendCommand(fmt.Sprintf("%s %s %s %v %v",
+		CmdSnapDestroy, in.Volume, in.Snapname, IoWaitTime, TotalWaitTime))
+	respstr := "ERR"
+	if len(sockresp) > 1 {
+		respstr = sockresp[1]
```
prateekpandey14: We are checking response length here.....and we are assuming if there is no error, `sockresp` length will be more then 1 ? if yes please add comments here at code level.

buildscripts/cstor-volume-mgmt/entrypoint.sh
```diff
@@ -2,6 +2,5 @@
 
 set -ex
 
-exec /usr/local/bin/cstor-volume-mgmt start
-exec service ssh start
-exec service rsyslog start
+/usr/local/bin/cstor-volume-mgmt start &
+exec /usr/local/bin/cstor-volume-grpc server start
```
prateekpandey14: nit : exec should be at start of the command line ?

Moteesh: if we put exec for the first line, the shell process will be replaced by the process that we invoked. it will not execute the following commands. that is why i used exec in the last line only.

cmd/cstor-volume-grpc/api/handler.go
Moteesh: if there is an error in writing to the UNIX socket, the sockresp will be nil. I need to check for nil here.

prateekpandey14: cool, can we add the code level comment here for better understanding and unit test

cmd/cstor-volume-grpc/api/handler_test.go
Moteesh: right now, we do not have any negative test cases to cover. All the errors are from istgt only. we will add it later if we need any.

cmd/cstor-volume-grpc/api/handler_test.go
Moteesh: right now, we do not have any negative test cases to cover. All the errors are from istgt only. we will add it later if we need any.

cmd/cstor-volume-grpc/app/command/commands.go
```diff
@@ -0,0 +1,289 @@
...
+	return nil
+}
+
+//CreateSnapshot creates snapshots
+func CreateSnapshot(volName, snapName string) (*v1alpha1.VolumeSnapCreateResponse, error) {
+	var conn *grpc.ClientConn
+	conn, err := grpc.Dial(fmt.Sprintf(":%d", api.VolumeGrpcListenPort), grpc.WithInsecure())
+	if err != nil {
+		glog.Fatalf("did not connect: %s", err)
+	}
```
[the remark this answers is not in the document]
Moteesh: done

PR #574, opened 2018-10-02
Fixes go lint in signal package

ChrsMark:
This PR closes https://github.com/openebs/openebs/issues/1964


ChrsMark:
@Akash4927, fixed that

pkg/signals/signal_posix.go
```diff
@@ -1,6 +1,6 @@
 // +build !windows
 
-/*
+/*Package signals provides basic signal's handling
```
sonasingh46: Can we remove the added comment here...
And just add a blank line between the comments and package declaration. 
I mean add a new line between line 17 and 18.
Thats all!

PR #593, opened 2018-10-03
[#1583] Add tests for /maya/pkg/task/identity.go 

RajatVaryani:
Signed-off-by: Rajat Varyani <rajat.varyani@outlook.com>

To improve the test coverage.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: 
fixes https://github.com/openebs/openebs/issues/1583

The intention is to get early feedback if you want to change something in the style of writing tests. Please provide feedback so that I can write tests for remaining functions. Thanks.


pkg/task/identity_test.go
```diff
@@ -24,7 +24,36 @@ import (
...
-func TestIsPod(t *testing.T) {}
+func TestIsPod(t *testing.T) {
+	tests := map[string]struct {
+		taskIdentifier taskIdentifier
+		isPod          bool
+	}{
+		"True": {
+			taskIdentifier: taskIdentifier{
+				identity: MetaTaskIdentity{
+					Kind: "Pod",
```
sonasingh46: You can also put a test case having `Pod` in all lower case to indicate that the comparison is case sensitive or not.

pkg/task/identity_test.go
```diff
@@ -24,7 +24,36 @@ import (
 func TestNewTaskIdentifier(t *testing.T) {}
 
 // TODO
-func TestIsPod(t *testing.T) {}
+func TestIsPod(t *testing.T) {
+	tests := map[string]struct {
+		taskIdentifier taskIdentifier
+		isPod          bool
+	}{
+		"True": {
```
sonasingh46: Instead of putting `True` as a key.
It would make more sense to put key as for pod `kindPod#1` and for not pod  `kindNotPod#1`, `kindNotPod#2` and so on.

RajatVaryani: Point taken.

pkg/task/identity_test.go
RajatVaryani: Agreed.

sonasingh46:
@RajatVaryani 
Can you please incorporate the requested changes?

RajatVaryani:
@sonasingh46 I have already. I am adding tests for other functions as well. It will be done by tonight.

RajatVaryani:
I have added tests for some functions. I will add for other functions as well.

pkg/task/identity_test.go
```diff
@@ -23,26 +23,354 @@ import (
...
-func TestIsPod(t *testing.T) {}
+func TestIsCommand(t *testing.T) {
+	tests := map[string]struct {
+		taskIdentifier taskIdentifier
+		isCommand      bool
+	}{
+		"kindCommand#1": {
+			taskIdentifier: taskIdentifier{
+				identity: MetaTaskIdentity{
+					Kind: "Command",
```
sonasingh46: What about adding a test case e.g. `Commands` ( to check for plural kind: whether `Commands` is a valid command type or not?)
Similarly for all other test functions below.

RajatVaryani:
The branch structure has screwed up. I will raise another PR.

PR #594, opened 2018-10-04
Fixing the formatting issues

Pensu:
Fixes https://github.com/openebs/openebs/issues/1973

There were some other formatting issues other than the ones in pkg/storagepool/storagepool_template_engine.go, this PR fixes them as well.

### DIFF
--- a/cmd/maya-agent/types/v1/storagebackend.go
+++ b/cmd/maya-agent/types/v1/storagebackend.go
@@ -7,7 +7,7 @@ type StorageBackendAdaptor struct {
 	MaxAllocationSize string `json:"maxAllocationSize"`
 }
 
-// HostDirSBA contains the generic options that apply to all the 
+// HostDirSBA contains the generic options that apply to all the
 // HostDirStorageBackendAdaptors (K8s CRD yaml)
 type HostDirSBA struct {
 	StorageBackendAdaptor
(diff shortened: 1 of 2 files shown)


pkg/storagepool/storagepool_template_engine.go
```diff
@@ -19,11 +19,11 @@ package storagepool
 
 import (
 	"fmt"
-	"strings"
 	"github.com/ghodss/yaml"
 	"github.com/openebs/maya/pkg/apis/openebs.io/v1alpha1"
 	"github.com/openebs/maya/pkg/engine"
 	"github.com/openebs/maya/pkg/util"
+	"strings"
 )
```
prateekpandey14: Same as above comment

Pensu: @prateekpandey14 I just ran ```go fmt``` on the said file and got the above result. I don't think there would be any other change in that particular file in terms of formatting, I will revert these changes, but then is the original issue valid?

Pensu: So I ran ```goimport``` and revert the changes, but now Better Code Hub is panicking, any idea why?

Pensu: Ok, got it, doesn't look like the files need any changes, closing this PR. Apologies for the noise. Probably the original issue should be closed as well.

Issue #629, opened 2018-10-06
Lint issue: types.go


Rishik233:
path: maya/cmd/maya-agent/k8s/types.go
Line 16: warning: exported type StorageBackendAdaptorSpec should have comment or be unexported (golint)
Line 23: warning: exported type StorageBackendAdaptorList should have comment or be unexported (golint)

Rishik233:
Working on this

princerachit:
@Rishik233 go ahead!

PR #633, opened 2018-10-06
fix lint issue in types.go

Rishik233:
fix lint issue in types.go
fixes https://github.com/openebs/maya/issues/629
Signed-off-by: rishik233 <rishik.rohan09@gmail.com>

### DIFF
--- a/cmd/maya-agent/k8s/types.go
+++ b/cmd/maya-agent/k8s/types.go
@@ -13,6 +13,8 @@ import (
 
 // #############################################################################
 
+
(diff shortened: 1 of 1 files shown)


cmd/maya-agent/k8s/types.go
```diff
@@ -13,13 +13,16 @@ import (
 
 // #############################################################################
 
+
```
sonasingh46: Just a minor pick,we can probably remove this extra line.

Rishik233: 👍 Would do that. Thanks!

PR #648, opened 2018-10-12
[TA3375] feat(clone): Support cstor snapshot clone

princerachit:
Support cStor volume clone from snapshot.

Update the snapshot name.
Remove unused fields from cstorvolumereplica.


cmd/cstor-pool-mgmt/volumereplica/volumereplica.go
```diff
@@ -107,7 +144,7 @@ func GetVolumes() ([]string, error) {
 
 // DeleteVolume deletes the specified volume.
 func DeleteVolume(fullVolName string) error {
-	deleteVolStr := []string{"destroy", "-r", fullVolName}
+	deleteVolStr := []string{"destroy", "-R", fullVolName}
```
princerachit: After discussion with @avishnu it was concluded that this change should be removed.

PR #652, opened 2018-10-13
Fixing golint issues in doc.go

Pensu:
Signed-off-by: Peeyush Gupta <peeyushgupta91@gmail.com>

Fixes: https://github.com/openebs/openebs/issues/2115


Pensu:
@princerachit I have a question, there are a lot of doc.go that have package v1 definition. Is it okay to comment every function with `Package v1 ...` or should there be a way to differentiate between different packages?

PR #679, opened 2018-10-13
added function comments #2152

Tushcr7:
Added comments to each functions which didn't have one using my understanding of the file, making it easier for documentation.

### DIFF
--- a/pkg/k8s/from_yaml.go
+++ b/pkg/k8s/from_yaml.go
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+//Package k8s
 package k8s
 
 import (
@@ -35,6 +36,7 @@ type DeploymentYml struct {
 	YmlInBytes []byte
 }
 
+// NewDeploymentYml returns a K8s deployment in case there is no error, else returns nil value
 func NewDeploymentYml(context, yml string, values map[string]interface{}) (*DeploymentYml, error) {
 	b, err := template.AsTemplatedBytes(context, yml, values)
 	if err != nil {
@@ -85,6 +87,7 @@ type ServiceYml struct {
 	YmlInBytes []byte
 }
 
+// NewServiceYml return a K8s service object in case there is no error, else returns nil
 func NewServiceYml(context, yml string, values map[string]interface{}) (*ServiceYml, error) {
 	b, err := template.AsTemplatedBytes(context, yml, values)
 	if err != nil {
@@ -133,6 +136,7 @@ type StoragePoolYml struct {
 	YmlInBytes []byte
 }
 
+// NewCStorPoolYml return a CStorPool object in case there is no error, else returns nil
 func NewCStorPoolYml(context, yml string, values map[string]interface{}) (*CStorPoolYml, error) {
 	b, err := template.AsTemplatedBytes(context, yml, values)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/k8s/from_yaml.go
```diff
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+//Package k8s
```
sonasingh46: Please remove this comment. Keeping a blank line at line 17 will do fine.

pkg/k8s/from_yaml.go
```diff
@@ -35,6 +36,7 @@ type DeploymentYml struct {
 	YmlInBytes []byte
 }
 
+// NewDeploymentYml returns a K8s deployment in case there is no error, else returns nil value
```
sonasingh46: `...returns a K8s deployment yaml....`

sonasingh46: Same can be done at all other places.
Also let us take view from others.
@princerachit @AmitKumarDas ( Can you please share some insights)

pkg/k8s/from_yaml.go
Tushcr7: Hi.
I have removed that comment. Please check now

pkg/apis/openebs.io/v1alpha1/cas_template_keys.go
```diff
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package v1alpha1
```
princerachit: This seems meaningless. If we cannot add any info let us remove this.

pkg/client/jiva/model.go
```diff
@@ -21,24 +21,28 @@ const (
 	snapSuffix         = ".img"
 )
 
+// Volumes
```
princerachit: There is no description ?!

pkg/apis/openebs.io/v1alpha1/cas_template_keys.go
Tushcr7: Yes
I had removed it initially. It was a mistake on my side. But the Codecov was giving a negative feedback. So added it back. But the status didn't change. I will remove it.

pkg/client/jiva/model.go
Tushcr7: Yeah. So I wasn't really sure about the description part and didn't want to add wrong description. So avoided it.
I'll try to go through the code again and add descriptions

pkg/apis/openebs.io/v1alpha1/cas_template_keys.go
sonasingh46: @Tushcr7 
You can remove the comment and put a blank line instead.

types/v1/suffix.go
```diff
@@ -14,6 +14,7 @@ limitations under the License.
 */
 
 // This is extracted from k8s.io/apimachinery/pkg/api/resource/suffix.go
+//Package v1
```
sonasingh46: We can remove this comment and put a blank line.

pkg/k8s/from_yaml.go
princerachit: Makes sense @sonasingh46

sonasingh46:
@Tushcr7 
Closing this PR due to inactivity. Please let us know if we need to work on the same.

PR #733, opened 2018-10-20
Simplify condition using increment and decrement

disc:
Fixes https://github.com/openebs/openebs/issues/2198
Golint fix: Simplify condition using increment and decrement


disc:
Added `Signed-off` to commit message

PR #783, opened 2018-11-09
feat(cstor-volume): Add replica status of a cstor volume.

princerachit:
Added a property ReplicaStatuses to cstorvolume which will store a list of statuses of
each replica connected to the target.

Generated/modified files in cstorvolume.pb.go and
zz_generated.deepcopy.go.

Restructured if-else block in handler.

The status field modified to:
```yaml
Status:
  Phase:  ReadOnly
  Replica Statuses:
    GUID:    16813495350490020069
    Status:  Degraded
    GUID:    11503620786459665966
    Status:  Degraded
    GUID:    13618884981032939368
    Status:  Running

```


princerachit:
There is an inherent issue with this PR, The replica name which is a big negative number overflows the current type casting. Therefore the replica names appearing in the cstor volume cr could be incorrect. This issue would be fixed in next commit. Trying to find a good conversion mechanism.

PR #820, opened 2018-11-21
[WIP] feat(api): enhance meta-data endpoint to return status of apiserver

prateekpandey14:
This commit will enhance the `meta-data/instance-id` endpoint to return current status of maya-apiserver. if all the required resources are created then API will return `Completed`, if not then returns `InProgress`


kmova:
Discussed with @prateekpandey14 and @AmitKumarDas . Closing this in favor of https://github.com/openebs/maya/pull/797

Probably in the future, checks can be added to the OpenEBS Operator to ensure that required sub-components are actually running.

PR #858, opened 2018-12-04
fix(mayactl): correct the calculations for volume metrics

kmova:
Fixed the following to match with the metrics returned
from exporter:
- size parameters are already returned in GB
- fix a bug in throughput calculation
- convert latency from ns to ms


ashishranjan738:
LGTM

PR #862, opened 2018-12-05
[US3000][TA3959] feat(usage health) Send openebs periodic health events

vharsh:
It is required to know about running openebs instances around the world and collect **periodic** anonymous information from clusters on user's consent.

**Which issue this PR fixes**
improvement: https://github.com/openebs/openebs/issues/2257

[old]
_To test the things, I have set the time duration to every ONE MINUTE, this will be changed to every ONE DAY, after it is tested by Travis-CI and other instances at openebs.ci_
[current]

1. If a volume is provisioned without replica-count details and if a user consents on sharing usage data with us, the replica-count value on our analytics dashboard will show up as `3` on `volume-provision` event.
2. This won't happen on a `volume-deprovision` event because of `pkg/usage/usage.go` L#213 which will only set it on volume-create. 
I pushed the condition inside `pkg/usage` as I felt it doesn't fit directly into the core business logic of `m-apiserver`
3. `OPENEBS_IO_HEALTH_ANALYTICS_PERIOD` if defined with `time.ParseDuration(string)` compatible values will be picked up, if not defined, our default value, i.e. `24 * time.Hour`  will be picked up.
4. A new func in `pkg/env` which returns the _passed default values_ if the string length of the env-value of `OPENEBS_IO_ANALYTICS_PING_INTERVAL` is `zero`, this prevents excess nesting of if-else code.

### DIFF
--- a/pkg/usage/usage.go
+++ b/pkg/usage/usage.go
@@ -19,11 +19,6 @@ import (
 	k8sapi "github.com/openebs/maya/pkg/client/k8s/v1alpha1"
 )
 
-const (
-	// GAclientID is the unique code of OpenEBS project in Google Analytics
-	GAclientID = "UA-127388617-1"
-)
-
 // Usage struct represents all information about a usage metric sent to
 // Google Analytics with respect to the application
 type Usage struct {
@@ -178,8 +173,8 @@ func (u *Usage) SetValue(v int64) *Usage {
 func (u *Usage) Build() *Usage {
 	// Default ApplicationID for openebs project is OpenEBS
 	v := NewVersion()
-	v.getVersion()
-	u.SetApplicationID("OpenEBS").
+	v.getVersion(false)
+	u.SetApplicationID(AppName).
 		SetTrackingID(GAclientID).
 		SetClientID(v.id)
 	// TODO: Add condition for version over-ride
@@ -191,7 +186,7 @@ func (u *Usage) Build() *Usage {
 // for non install events
 func (u *Usage) ApplicationBuilder() *Usage {
 	v := NewVersion()
-	v.getVersion()
+	v.getVersion(false)
 	u.SetApplicationVersion(v.openebsVersion).
 		SetApplicationName(v.k8sArch).
 		SetApplicationInstallerID(v.k8sVersion).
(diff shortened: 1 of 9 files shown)


pkg/usage/usage.go
```diff
@@ -206,11 +206,25 @@ func (u *Usage) SetVolumeCapacity(volCapG string) *Usage {
...
+// Wrapper for setting replica count for volume events
+// NOTE: This doesn't get the replica count in a volume de-provision event.
+// TODO: Pick the current value of replica-count from the CAS-engine
+func (u *Usage) SetReplicaCount(count, method string) *Usage {
+	if method == "volume-provision" && count == "" {
+		// Case: When volume-provision the replica count isn't specified
+		// it is set to three by default by the m-apiserver
+		u.SetAction("replica:3")
+	}
+	u.SetAction("replica:" + count)
```
kmova: should there be an else here?

vharsh: Not unless we want to send something other than `replica:` for a `volume-deprovision` event.
Our limitation here is that we don't get the replica-count on volume-deprovision.

pkg/usage/usage.go
```diff
@@ -206,11 +206,25 @@ func (u *Usage) SetVolumeCapacity(volCapG string) *Usage {
 	return u
 }
 
+// Wrapper for setting replica count for volume events
+// NOTE: This doesn't get the replica count in a volume de-provision event.
+// TODO: Pick the current value of replica-count from the CAS-engine
+func (u *Usage) SetReplicaCount(count, method string) *Usage {
+	if method == "volume-provision" && count == "" {
```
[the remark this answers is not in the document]
vharsh: background info:
In the CASvolume struct the replica count is a `string`, so parsing it to an `int` and sending it to Google's `event-action` which accepts a string isn't a good idea as we are using `event-value` to record the `volume-size` which is an `int64`

pkg/usage/usage.go
vharsh: Okay, I get the issue now an else is required.

vharsh: Fixed in `35d26d`

PR #877, opened 2018-12-26
enhance(travis) Exit running tests when images aren't built

vharsh:
This PR is an alternative take to skip running integration tests when
    there are build/UT errors.
    
    It is possible that the image doesn't exist, in that case the `docker
    tag` command will fail and return a non-zero return code, which will
    stop the build.
    
    fixes: https://github.com/openebs/openebs/issues/2320
    
    Signed-off-by: Harsh Vardhan <harsh.vardhan@mayadata.io>

This change is being done to avoid running e2e tests with `minikube` in Travis-CI when some UT fails.
The same could also be achieved by setting `imagePullPolicy` to `Never` but this would be a bit unreadable(as we'd still have to pull Node-disk-manager), i.e. only first four occurrence of five.

**Which issue this PR fixes** 
https://github.com/openebs/openebs/issues/2320

I have added a second commit which I'll be using to test the Travis changes, this will be dropped after it has been verified.

from 
```bash
$ help set
```
`  -e  Exit immediately if a command exits with a non-zero status.`


vharsh:
Travis logs from `vharsh/maya`
```
0.29s$ ./ci/travis-ci.sh
*****************************Retagging images and setting up env***************************
Error response from daemon: No such image: openebs/m-apiserver:ci
The command "./ci/travis-ci.sh" exited with 1.
```

princerachit:
What if some other command fails like installing or getting iscsi/(d) status etc? Would this behaviour be accepted? @vharsh

vharsh:
@princerachit Setting `set -e` will immediately stop the build if any of the shell commands fail with a non-zero exit code. If that has to be avoided a `set +e` can be used to re-instate previous behaviour.

kmova:
@vharsh - can you please take a look at the travis failure?

vharsh:
@kmova I have added a bug in one of the unit-tests to verify that this will actually work, specified in `Special notes for the reviewer`.
Besides should the scripts also exit if other things report an error as @princerachit has stated?

vharsh:
@prateekpandey14 
> Setting set -e will immediately stop the build if any of the shell commands fail with a non-zero exit code. If that has to be avoided a set +e can be used to re-instate previous behaviour.

Is this required?

PR #894, opened 2019-01-17
feat(CAS_Template)add liveness for cstor pool

sonasingh46:
This PR will add liveness check for cstor-pool container.
There are scenarios where I/O can be hung and in that case we may need to restart the cstor-pool container.

### DIFF
--- a/pkg/install/v1alpha1/cstor_pool.go
+++ b/pkg/install/v1alpha1/cstor_pool.go
@@ -215,6 +215,16 @@ spec:
               protocol: TCP
             - containerPort: 3232
               protocol: TCP
+            livenessProbe:
+              exec:
+                command:
+                - /bin/sh
+                - -c
+                - zfs set io.openebs:livenesstimestap='$(date)' cstor-$OPENEBS_IO_CSTOR_ID
+              failureThreshold: 3
+              initialDelaySeconds: 300
+              periodSeconds: 10
+              timeoutSeconds: 30
             securityContext:
               privileged: true
             volumeMounts:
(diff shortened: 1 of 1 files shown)


sonasingh46:
@vishnuitta 
We need POOL_UID to get the pool name. This UID is exposed in mgmt container and not in the pool container. 
So we need to expose it there also.

vishnuitta:
> @vishnuitta
> We need POOL_UID to get the pool name. This UID is exposed in mgmt container and not in the pool container.
> So we need to expose it there also.

Got it. Want experts review on this

sonasingh46:
Sure @vishnuitta

pkg/install/v1alpha1/cstor_pool.go
```diff
@@ -215,6 +215,16 @@ spec:
...
+            livenessProbe:
+              exec:
+                command:
+                - /bin/sh
+                - -c
+                - zfs set io.openebs:livenesstimestap='$(date)' cstor-$OPENEBS_IO_CSTOR_ID
+              failureThreshold: 3
+              initialDelaySeconds: 300
+              periodSeconds: 10
+              timeoutSeconds: 30
```
avishnu: Is there a possibility of a clash between periodSeconds and timeoutSeconds? For instance, the current probe is not yet timed-out and the next one has started.

sonasingh46: No. Until the probe finishes either with a success or failure case next probe should not start.

vishnuitta: as per the doc, it starts another probe @avishnu as you expected. After `timeoutSeconds`, if no response comes, another probe gets started after 10 seconds (i.e., 40 seconds after first probe). This happens for 3 times, and, if the 3rd one also fails, pod gets restarted. So, you can assume - (10+30)*3 seconds = 120 seconds to identify for pod to be marked as bad in case of any IO hang.

avishnu: Ok

PR #898, opened 2019-01-21
[US4156] feat(spc-watcher): Add support for raidz and raidz2

utkarshmani1997:

- To add support for raidz and raidz2 type

- Depends on #893


utkarshmani1997:
Closing since it will be part of #893

PR #902, opened 2019-01-23
fix(policies): support for resource requests on side-cars (#889)

kmova:
Ref: https://github.com/openebs/openebs/issues/2294

With 0.7.0, there was support for configuring resource
limits on the pool and volume pod side-cars using
AuxilaryResourceLimits. With this PR, added the support
to specify AuxilaryResourceRequests.

This feature is useful in cases where user has to specify
minimum requests like ephemeral storage etc. to avoid
erroneous eviction by K8s as described in the above issue.

Also, some of the resource types are very version dependent
so keeping this as generic and not providing any defaults in
this PR.

In the future, it is possible that OpenEBS Operator could
automatically determine the version and set default values
on the generated default Storage Class.

(cherry picked from commit 9f12a28fe7b755e21715fb627c89b0054c22e87f)

### DIFF
--- a/pkg/install/v1alpha1/cstor_pool.go
+++ b/pkg/install/v1alpha1/cstor_pool.go
@@ -63,6 +63,10 @@ spec:
   #      memory: 1Gi
   - name: PoolResourceLimits
     value: "none"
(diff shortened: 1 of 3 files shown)


vishnuitta:
but, travis is failing

PR #967, opened 2019-03-07
fix(webhook): ignore delete request for missing resource req name

prateekpandey14:
#### why we need this PR

This fixes a bug where deleting a namespace, that doesn't have any PVC resource can trigger this validating webhook. This nil PVC resulted into a runtime error thus invalidating the validation. As a result of this invalidation, the validation is being retried that and ends up being in the same state forever.

#### What this PR does 
Request name as nil will be ignored by the webhook validation policy. In other words webhook will validate this request successfully.

fixes: openebs/openebs/issues/2431

**Checklist:**
- [x] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -130,6 +130,12 @@ func (wh *webhook) validate(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionRespo
 		response.Allowed = true
 		return response
 	}
+	// ignore the Delete request if requested resource doesn't exists
+	if ar.Request.Name == "" {
+		response.Allowed = true
+		return response
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/webhook/webhook.go
```diff
@@ -130,6 +130,12 @@ func (wh *webhook) validate(ar *v1beta1.AdmissionReview) *v1beta1.AdmissionRespo
 		response.Allowed = true
 		return response
 	}
+	// ignore the Delete request if requested resource doesn't exists
```
vishnuitta: comment can be like
`ignore the Delete request of PVC if resource name is empty which can happen as part of cleanup process of namespace`

prateekpandey14: Updated the comments

PR #985, opened 2019-03-13
feat(castemplate) : add support for remote execute of a command in a container of a pod

imazik:

add support for remote execute of a command in a container of a pod

We can use castemplate to run command remotely this will return response which contains stdout and stderr. We can pass pod exec config in runtask. Here is an example
```
kind: RunTask
metadat:
  name: task-remoteexec
  namespace: default
spec:
  meta: |
    id: remoteexec
    apiVersion: v1
    kind: Pod
    action: exec
    objectName: ubuntu
    runNamespace: default
  post: |-
    '{{- jsonpath .JsonResult "{.Stdout}" | trim | saveAs "remoteexec.Stdout" .TaskResult | noop -}}'
    '{{- jsonpath .JsonResult "{.Stderr}" | trim | saveAs "remoteexec.Stderr" .TaskResult | noop -}}'
  task: |-
    command:
    - ps
    - aux
    container: ubuntu
    stdin: false
    stdout: true
    stderr: true
```

we can get output of remotely executed command in stdout and stderr.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests


pkg/task/ecec.go
```diff
@@ -0,0 +1,92 @@
+/*
```
AmitKumarDas: I would like to deviate from existing package structure.
Will it be possible to make use of idiomatic maya packaging? e.g. below comes to my mind
- verify the package path
- verify the naming
- verify the patterns

```go
// in pkg/kubernetes/podexec/v1alpha1/podexec.go

import (
  "github.com/pkg/errors"
)

type podexec struct {
  object             *api_core_v1.PodExecOptions
  ignoreErrors   bool
  errs                []err
}

func (p *podexec) AsAPIPodExec() (*api_core_v1.PodExecOptions, error) {
  err = p.Validate()
  if err != nil && !p.ignoreErrors {
    return nil, err
  }
  return p.object, nil
}

func WithTemplate(context, yaml string, values map[string]interface{}) (*podexec, error) {
  b, err := template.AsTemplatedBytes(context, yaml, values)
  if err != nil {
    return nil, err
  }
  var exec api_core_v1.PodExecOptions
  err = yaml.Unmarshal(b, &exec)
  if err != nil {
    return nil, err
  }
  p.object = exec
  return p, nil
}

func (p *podexec) Validate() error {
  if len(p.errs) != 0 {
    return errors.Errorf("validation failed: %v", p.errs)
  }
  if p.object.Command == "" {
    return errors.New("validation failed: command not provided")
  }
  if p.object.ContainerName == "" {
    return errors.New("validation failed: container name not provided")
  }
  return nil
}

type buildOption func(*podexec)

func IgnoreErrors() buildOption {
  return func(p *podexec) {
    p.ignoreErrors = true
  }
}

func (p *podexec) Apply(opts ...buildOption) (*podexec, error) {
  for _, o := range opts {
    o(p)
  }
  return p
}
```

pkg/task/ecec.go
```diff
@@ -0,0 +1,92 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package task
+
+import (
+	"errors"
```
AmitKumarDas: Avoid standard library `errors` package.
make use of `github.com/pkg/errors` everywhere

pkg/task/ecec.go
imazik: done please have a look

pkg/task/ecec.go
imazik: done

pkg/client/k8s/k8s.go
```diff
@@ -1128,6 +1132,63 @@ func (k *K8sClient) DeleteOEV1alpha1CVR(name string) error {
 	})
 }
 
+// ExecCoreV1Pod run a command remotely in a container of a pod
+func (k *K8sClient) ExecCoreV1Pod(podExecOptions *api_core_v1.PodExecOptions,
```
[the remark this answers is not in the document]
imazik: in method `ExecCoreV1Pod()` name argument should be pod name for other resources this is name only. I think it is more readable. Please have a look from caller else I will change it.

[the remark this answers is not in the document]
imazik: stdin is for runtime input of a command. I am planing to make it not configurable and put it nil.

pkg/client/k8s/k8s.go
```diff
@@ -1128,6 +1132,63 @@ func (k *K8sClient) DeleteOEV1alpha1CVR(name string) error {
...
+	name string, stdin io.Reader) (result []byte, err error) {
+
+	// create request object for exec
+	req := k.cs.CoreV1().RESTClient().Post().
+		Resource("pods").
+		Name(name).
+		Namespace(k.ns).
+		SubResource("exec")
+
+	scheme := runtime.NewScheme()
```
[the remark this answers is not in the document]
imazik: added. It is creating one empty schema and add it to request to get url values and param from req.
replaced this code with
```
req := k.cs.CoreV1().RESTClient().Post().
		Resource("pods").
		Name(podName).
		Namespace(k.ns).
		SubResource("exec").
		VersionedParams(podExecOptions, scheme.ParameterCodec)
``` 
empty schema is already present in `scheme.ParameterCodec`

pkg/client/k8s/k8s.go
```diff
@@ -1128,6 +1132,63 @@ func (k *K8sClient) DeleteOEV1alpha1CVR(name string) error {
...
+	// create exec executor which is an interface for transporting shell-style streams.
+	exec, err := remotecommand.NewSPDYExecutor(config, "POST", req.URL())
+	if err != nil {
+		return
+	}
+	var stdout, stderr bytes.Buffer
+	// Stream initiates the transport of the standard shell streams. It will transport any
+	// non-nil stream to a remote system, and return an error if a problem occurs.
+	err = exec.Stream(remotecommand.StreamOptions{
+		Stdin:  stdin,
```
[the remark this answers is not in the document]
imazik: done

pkg/client/k8s/k8s.go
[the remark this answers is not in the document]
AmitKumarDas: ok... perhaps order the arguments. first comes name then comes the podexecoptions.

pkg/kubernetes/podexec/v1alpha1/podexec.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, err
+	}
+	p.object = exec
+	return p, nil
+}
+
+// Validate validates PodExecOptions it mainly checks for container name is present or not and
+// commands are present or not.
+func (p *podexec) Validate() error {
+	if len(p.errs) != 0 {
```
AmitKumarDas: we can remove this property i.e. `errs` since we are not using it. we will revisit this when required.

pkg/kubernetes/podexec/v1alpha1/podexec.go
```diff
@@ -0,0 +1,89 @@
...
+
+// IgnoreErrors is a buildOption that is used ignore errors
+func IgnoreErrors() buildOption {
+	return func(p *podexec) {
+		p.ignoreErrors = true
+	}
+}
+
+// Apply applies all build options in podexec
+func (p *podexec) Apply(opts ...buildOption) (*podexec, error) {
```
AmitKumarDas: let us remove error from signature since its not being used.
we shall revisit when required.

pkg/kubernetes/podexec/v1alpha1/podexec.go
```diff
@@ -0,0 +1,89 @@
...
+	err = p.Validate()
+	if err != nil && !p.ignoreErrors {
+		return nil, err
+	}
+	return p.object, nil
+}
+
+// WithTemplate takes Yaml values which is given in runtask and key in which configuration
+// is present and unmarshal it with PodExecOptions.
+func WithTemplate(context, yamlString string, values map[string]interface{}) (p *podexec, err error) {
```
AmitKumarDas: i felt avoiding error from the signature would have made the calling code readable.
I felt using err as a value will be good.
in other words if err was captured into `errs` property; then calling code would look like:
```go
e, err := podexec.WithTemplate(abc, yml, nil).AsAPIPodExec()
// or 
e, err := podexec.WithTemplate(xyz, yaml, nil).Apply(myopts).AsAPIPodExec()
```

Above will not force calling code to keep checking for err everytime a method is invoked.
Assumption is some of the methods are used to build the object while some of the methods are invoked finally.
Here `AsAPIPodExec` is the final method.

However, this comment about error handling is optional & should be considered on a case-by-case basis.

pkg/task/task.go
```diff
@@ -26,6 +26,7 @@ import (
 
 	m_k8s_client "github.com/openebs/maya/pkg/client/k8s"
 	m_k8s "github.com/openebs/maya/pkg/k8s"
+	m_k8s_podexec "github.com/openebs/maya/pkg/kubernetes/podexec/v1alpha1"
```
AmitKumarDas: `podexec` will suffice as an alias

pkg/task/task.go
```diff
@@ -1053,6 +1056,28 @@ func (m *taskExecutor) deleteOEV1alpha1CSV() (err error) {
 	return
 }
 
+// execCoreV1Pod runs given command remotely in given container of given pod
+// and post stdout and and stderr in JsonResult. You can get it using -
+// {{- jsonpath .JsonResult "{.Stdout}" | trim | saveAs "XXX" .TaskResult | noop -}}
+func (m *taskExecutor) execCoreV1Pod() (err error) {
+	podexecoptsFromTemplate, err := m_k8s_podexec.WithTemplate("execCoreV1Pod", m.runtask.Spec.Task, m.templateValues)
```
AmitKumarDas: the names used here can be single character names. e.g.:
```go
e, err := podexec.WithTemplate("taskname", taskYaml, values)
if err != nil {
  return err
}
podExec, err := e.AsAPIPodExec()
if err != nil {
  return err
}
```
Above does not hinder the readability IMO

Alternatively, if we handle error as a value this can become:
```go
execOpts, err := podexec.
       WithTemplate("taskname", taskYaml, values).
       AsAPIPodExec()
if err != nil {
  return err
}
```

imazik: done

pkg/task/task.go
imazik: done

pkg/kubernetes/podexec/v1alpha1/podexec.go
imazik: done

pkg/kubernetes/podexec/v1alpha1/podexec.go
imazik: done

pkg/client/k8s/k8s.go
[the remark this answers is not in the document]
imazik: done

PR #1026, opened 2019-03-29
Refact(Version): bump version to 0.8.2

prateekpandey14:


vishnuitta:
I see changes in runTasks also as done in this other similar PR https://github.com/openebs/maya/pull/794.. do we need to do that here as welll?

prateekpandey14:
@vishnuitta not required as from 0.8.1 we suffix the version at runtime by reading from `VERSION` file

PR #1066, opened 2019-04-15
feat(install, cas templates): adds openebs.io/version label to the OpenEBS dataplane components

mittachaitu:

**What this PR does**:
This PR adds the openebs.io/version label to the data plane pods.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #https://github.com/openebs/openebs/issues/2516

**Checklist:**
- [x] Fixes #openebs/openebs#2516
- [ ] Commit has unit tests

### DIFF
--- a/pkg/install/v1alpha1/cstor_pool.go
+++ b/pkg/install/v1alpha1/cstor_pool.go
@@ -53,7 +53,7 @@ spec:
     value: {{env "OPENEBS_SERVICE_ACCOUNT"}}
   # PoolResourceRequests allow you to specify resource requests that need to be available
   # before scheduling the containers. If not specified, the default is to use the limits
-  # from PoolResourceLimits or the default requests set in the cluster. 
+  # from PoolResourceLimits or the default requests set in the cluster.
   - name: PoolResourceRequests
     value: "none"
   # PoolResourceLimits allow you to set the limits on memory and cpu for pool pods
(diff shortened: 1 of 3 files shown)


pkg/install/v1alpha1/cstor_pool.go
```diff
@@ -200,11 +200,13 @@ spec:
       selector:
         matchLabels:
           app: cstor-pool
+          openebs.io/version: {{ .CAST.version }}
```
vishnuitta: why do we need here when version is already part of `metadata.labels`?

sonasingh46: @vishnuitta -- labels present in `matchLabels` are those which should be matched with labels present in metadata for a controller e.g. deployment to manage the pods.
Now we need to see do we also need the version labels as part of this matching.

mittachaitu: Reason to add openebs.io/version labels under `matchLabels` is 1) Deployment controller will first delete the (rs, pods) which doesn't have the matching label in `.spec.template.metadata.labels`(useful in update scenarios). 
2) Deployment controller will never allow to spawn the pods if matching label value is not found in `template.metadata.labels`.

vishnuitta:
setting version need update for upgrade scripts as well.

mittachaitu:
> setting version need update for upgrade scripts as well.

I labeled it with `need-upgrade`

mittachaitu:
> setting version need update for upgrade scripts as well.

I labeled it with `need-upgrade`

PR #1177, opened 2019-05-10
feat(localpv): auto-install localpv hostpath storage class

kmova:
This PR automates the installation of the Local PV
StorageClass for StorageType=hostpath, as part of the
maya-apiserver startup.

### DIFF
--- a/pkg/install/v1alpha1/localpv_sc.go
+++ b/pkg/install/v1alpha1/localpv_sc.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2019 The OpenEBS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


pkg/install/v1alpha1/localpv_sc.go
```diff
@@ -0,0 +1,61 @@
...
+// TODO
+// Rename this file by removing the version suffix information
+package v1alpha1
+
+const localPVSCYamls = `
+---
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: hostpath
```
prateekpandey14: sc name `hostpath` is generic to k8s, can we name it as  `openebs-hostpath` ? .

kmova: good point! changed it.

PR #1187, opened 2019-05-13
fix(snapshots): fix jiva snaps with pods in openebs ns

kmova:
This PR introduces a new  runtask to fetch the target service
using the values provided by the snap cast engine.

Also fixes the travis tests to use the locally build images instead of 
pulling in from docker hub. 

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/install/v1alpha1/jiva_snapshot.go
+++ b/pkg/install/v1alpha1/jiva_snapshot.go
@@ -31,7 +31,7 @@ spec:
   taskNamespace: {{env "OPENEBS_NAMESPACE"}}
   run:
     tasks:
-    - jiva-volume-podsinopenebsns-default
+    - jiva-snapshot-podsinopenebsns-default
     - jiva-snapshot-create-listsourcetargetservice-default
     - jiva-snapshot-create-invokehttp-default
   output: jiva-snapshot-create-output-default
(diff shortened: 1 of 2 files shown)


pkg/install/v1alpha1/jiva_snapshot.go
```diff
@@ -31,18 +31,34 @@ spec:
...
-    - jiva-volume-podsinopenebsns-default
+    - jiva-snapshot-podsinopenebsns-default
     - jiva-snapshot-create-listsourcetargetservice-default
     - jiva-snapshot-create-invokehttp-default
   output: jiva-snapshot-create-output-default
 ---
 apiVersion: openebs.io/v1alpha1
 kind: RunTask
+metadata:
+  name: jiva-snapshot-podsinopenebsns-default
```
AmitKumarDas: Do we intend to use it as a common runtask?
Why is the name of this runtask different than others?

prateekpandey14: Yes ... that was the mistake , we were using `.Volume` property for snapshot related operations as well. and that was breaking the runtask .

pkg/install/v1alpha1/jiva_snapshot.go
```diff
@@ -31,18 +31,34 @@ spec:
...
+  meta: |
+    id: jivasnappodsinopenebsns
+    runNamespace: {{ .Config.OpenEBSNamespace.value }}
+    apiVersion: v1
+    kind: Service
+    action: list
+    options: |-
+      labelSelector: openebs.io/controller-service=jiva-controller-svc,openebs.io/persistent-volume={{ .Snapshot.volumeName }}
+  post: |
+    {{- jsonpath .JsonResult "{.items[*].metadata.namespace}" | trim | saveAs "jivasnappodsinopenebsns.ns" .TaskResult | noop -}}
```
utkarshmani1997: listing a service require namespace also, by default it will search in default namespace with the given labels. @AmitKumarDas how runtask process this post request ? does this list in runNamespace ?

kmova: yes @utkarshmani1997 .. it uses the runNamespace.

pkg/install/v1alpha1/jiva_snapshot.go
kmova: the snapshot and the volume requests are using different cast engines and the values are passed in .Snapshot and .Volume objects.

pkg/install/v1alpha1/jiva_snapshot.go
utkarshmani1997: for exp:
image

pkg/install/v1alpha1/jiva_snapshot.go
AmitKumarDas: can this RunTask be named with `jiva-snapshot-create-`

pkg/install/v1alpha1/jiva_snapshot.go
utkarshmani1997: @kmova so, if user will deploy in "default" namespace, what will be value of "runNamespace", will it be overridden to "default" ?

PR #1326, opened 2019-07-02
add new cstor pool schema and autogen code

sonasingh46:
Add new CSP schema and autogen code


pkg/apis/openebs.io/v1alpha1/new_cstor_pool.go
```diff
@@ -0,0 +1,59 @@
...
+type NewTestCStorPool struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   NewCStorPoolSpec `json:"spec"`
+	Status CStorPoolStatus  `json:"status"`
+}
+
+// CStorPoolSpec is the spec listing fields for a CStorPool resource.
+type NewCStorPoolSpec struct {
```
mynktl: Instead of creating new one, can't we use `PoolSpec` from cstorpoolcluster?

pkg/apis/openebs.io/v1alpha1/new_cstor_pool.go
```diff
@@ -0,0 +1,59 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +resource:path=cstorpool
+
+// NewTestCStorPool describes a cstor pool resource created as custom resource.
+type NewTestCStorPool struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   NewCStorPoolSpec `json:"spec"`
+	Status CStorPoolStatus  `json:"status"`
```
mynktl: can we use spec and status from cstorpoolcluster here?

sonasingh46: Spec cannot be entirely reused from cstorpoolcluster.
we can think of reusing status. We will enhance as we go ahead in the PR.

pkg/apis/openebs.io/v1alpha1/new_cstor_pool.go
sonasingh46: It will conflict..

mynktl: Ok

pkg/apis/openebs.io/v1alpha1/new_cstor_pool.go
mynktl: ok

PR #1339, opened 2019-07-10
chore(cstor-operator): fix BD claim, pool resource naming and creation and node reprocessing

sonasingh46:
This commit will do following:

1. Once BDC has been created for BD ( for pool provisioning), a CSP
   will not be created once all the BDs are in *Claimed* state.
   This means, more then one event will be rquired for a CSP to be created.
   For the first time during pool provisioning, BD will be cheked for Claimed
   state and if not claimed, BDC for them will be created and program shall
   return.
   In the next event, again the BDs will be checked for claimed state and if
   claimed pool provisioning shall proceed else program shall return and so on.

2. In a single event, when a node has been visited for pool creation, in either
   case success or failure, that node will not be revisited for pool creation.
   This is required as one dirty node causing pool provisioning failure can stop
   other subsequent nodes from being pool provisioned on them.

3. Pool resources i.e. CSP and Pool Dpeloyment will have names that is suffixed by 
    char alphanumeric random string.

4. Pool deployment build was failing due to missing namespace. This PR contains a fix
    for this.

### DIFF
--- a/pkg/algorithm/nodeselect/v1alpha2/select_node.go
+++ b/pkg/algorithm/nodeselect/v1alpha2/select_node.go
@@ -40,14 +40,15 @@ func (ac *Config) SelectNode() (*apis.PoolSpec, string, error) {
 		// pin it
 		pool := pool
 		nodeName, err := GetNodeFromLabelSelector(pool.NodeSelector)
+		if err != nil || nodeName == "" {
+			glog.Errorf("could not use node for selectors {%v}", pool.NodeSelector)
+			continue
+		}
 		if ac.VisitedNodes[nodeName] {
 			continue
 		} else {
 			ac.VisitedNodes[nodeName] = true
-			if err != nil {
-				glog.Errorf("could not use node for selectors {%v}", pool.NodeSelector)
-				continue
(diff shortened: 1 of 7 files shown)


pkg/algorithm/nodeselect/v1alpha2/select_node.go
```diff
@@ -40,13 +40,20 @@ func (ac *Config) SelectNode() (*apis.PoolSpec, string, error) {
 		// pin it
 		pool := pool
 		nodeName, err := GetNodeFromLabelSelector(pool.NodeSelector)
-		if err != nil {
+		if err != nil || nodeName == "" {
```
mittachaitu: nit: IMO `nodeName == ""` is not required

sonasingh46: Just added an extra safe check after having discussion with @vishnuitta

vishnuitta: added for sake of safety in future for any code change in GetNodeFromLabelSelector.. code maintainability

PR #1347, opened 2019-07-15
chore(cspc, cstor-operator): record events on cspc

sonasingh46:
This PR does the following :
1. Records events on CSPC.


cmd/maya-apiserver/cstor-operator/cspc/handler.go
```diff
@@ -122,26 +122,41 @@ func (c *Controller) syncCSPC(cspc *apis.CStorPoolCluster) error {
 	//}
 	openebsNameSpace := env.Get(env.OpenEBSNamespace)
 	if openebsNameSpace == "" {
+
+		message := fmt.Sprint("Could not sync CSPC: got empty namespace for openebs from env variable")
+		c.recorder.Event(cspc, apis.EventTypeError, "Getting Namespace", message)
 		return errors.Errorf("Could not sync CSPC {%s}: got empty namespace for openebs from env variable", cspc.Name)
```
vishnuitta: returning 'nil' is better to avoid requeueing

vishnuitta: my take on this is: returning nil on all the cases except when we know that faster reconcilation helps.
one such example is: if BDCs are just created, I expect faster reconcilation (may be with induced delay like for a second)

sonasingh46: Ok. Will fix it.

mynktl: What should be the ideal behavior if `OPENEBS_NAMESPACE` is not set? We are setting `OPENEBS_NAMESPACE` through operator only. If it is not set then there is a problem in deployment. So we should not start Maya-apiserver if it is not set. Or we should set from `maya-apiserver` itself?

cmd/maya-apiserver/cstor-operator/cspc/runner.go
```diff
@@ -95,26 +104,43 @@ func (c *Controller) processNextWorkItem() bool {
...
 		// get queued again until another change happens.
 		c.workqueue.Forget(obj)
 		glog.V(1).Infof("Successfully synced '%s'", key)
-		return nil
+		return nil, key
 	}(obj)
 
 	if err != nil {
-		runtime.HandleError(err)
+		c.handleErr(err, key)
```
mynktl: Let's avoid multiple checks for `err`. `handleErr` is already checking if `err` is `nil` or not.

cmd/maya-apiserver/cstor-operator/cspc/runner.go
```diff
@@ -95,26 +104,43 @@ func (c *Controller) processNextWorkItem() bool {
...
+
+func (c *Controller) handleErr(err error, key interface{}) {
+	if err == nil {
+		c.workqueue.Forget(key)
+		return
+	}
+
+	if c.workqueue.NumRequeues(key) < maxRetries {
+		glog.V(2).Infof("Requeuing CSPC as sync failed %v: %v", key, err)
+		c.workqueue.AddRateLimited(key)
```
mynktl: Why do we need to re-queue? We are already queuing on a periodic `update` event.

pkg/apis/openebs.io/v1alpha1/event_types.go
```diff
@@ -0,0 +1,27 @@
...
+package v1alpha1
+
+// Valid values for event types (new types could be added in future)
+const (
+	// EventTypeNormal is for information only and will not cause any problems
+	EventTypeNormal string = "Normal"
+	// EventTypeWarning is to warn that something might go wrong
+	EventTypeWarning string = "Warning"
+	// EventTypeError is to indicate that something went wrong
+	EventTypeError string = "Error"
```
mynktl: Nice!

cmd/maya-apiserver/cstor-operator/cspc/handler.go
```diff
@@ -122,26 +122,41 @@ func (c *Controller) syncCSPC(cspc *apis.CStorPoolCluster) error {
...
+		c.recorder.Event(cspc, apis.EventTypeError, "Getting Pending Pool(s) ", message)
+		return errors.Wrapf(err, "Could not sync CSPC {%s}: failed to get pending pool count", cspc.Name)
 	}
+
 	if pendingPoolCount > 0 {
 		err = pc.create(pendingPoolCount, cspc)
 		if err != nil {
-			return err
+			message := fmt.Sprintf("Could not create pool(s) for CSPC: %s", err.Error())
+			c.recorder.Event(cspc, apis.EventTypeError, "Pool Create", message)
```
mittachaitu: Here currently we are getting `error only if we are not able to acquire lease` is it really necessary to keep in front of users about lease errors?

sonasingh46: I think so. Else how will we know what caused the failure.

cmd/maya-apiserver/cstor-operator/cspc/runner.go
sonasingh46: We are queuing it so that we do not wait till the next update event.

mynktl: ok

sonasingh46: But anyway -- removed that backoff code.

cmd/maya-apiserver/cstor-operator/cspc/runner.go
sonasingh46: done

cmd/maya-apiserver/cstor-operator/cspc/handler.go
```diff
@@ -162,9 +181,15 @@ func (pc *PoolConfig) create(pendingPoolCount int, cspc *apis.CStorPoolCluster)
 	glog.V(4).Infof("Lease acquired successfully on cstorpoolcluster %s ", cspc.Name)
 	defer newSpcLease.Release()
 	for poolCount := 1; poolCount <= pendingPoolCount; poolCount++ {
+
+		message := fmt.Sprintf("Provisioning pool %d/%d ", poolCount, pendingPoolCount)
+		pc.Controller.recorder.Event(cspc, "Info", "Create", message)
```
vishnuitta: instead of this, how about raising event for success case.. same goes for log in next line

cmd/maya-apiserver/cstor-operator/cspc/storagepool_create.go
```diff
@@ -79,10 +79,15 @@ func (pc *PoolConfig) CreateStoragePool() error {
 	gotCSP, err := pc.createCSP(cspObj)
 
 	if err != nil {
-		return errors.Wrapf(err, "failed to create csp for cspc {%s}", pc.AlgorithmConfig.CSPC.Name)
+		return errors.New("failed to create CSP")
```
vishnuitta: better to have old error also?
and name/nodename of cspObj also would be good

PR #1375, opened 2019-07-25
[WIP]fix(cstor-pool): delete sock file before deleting for cstor-pool container

prateekpandey14:

This PR deletes the unix.sock file (which is shared between the
cstor-pool-mgmt and cstor-pool). This is deleted during cstor-pool container
deletion phase via a preStop hook.

(https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/)
In other words, before triggering kill of cstor-pool container, kubernetes will
execute this preStop hook.

This fix avoids RPC between cstor-pool-mgmt container of a new pod and
cstor-pool container of the terminating pod i.e. old pod.

### DIFF
--- a/pkg/install/v1alpha1/cstor_pool.go
+++ b/pkg/install/v1alpha1/cstor_pool.go
@@ -280,6 +280,13 @@ spec:
               postStart:
                  exec:
                     command: ["/bin/sh", "-c", "sleep 2"]
+              # Removes the sock file that is used for communication between
+              # cstor-pool and cstor-pool-mgmt containers, so that, cstor-pool-mgmt
+              # of new pool pod instance will NOT communicate with cstor-pool
+              # container of terminating pod instance.
+              preStop:
+                 exec:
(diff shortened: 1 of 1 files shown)


sonasingh46:
As discussed we need a revisit with and look into other possible solutions e.g. a directory with pod uid name where sock while will be created.

pkg/install/v1alpha1/cstor_pool.go
```diff
@@ -280,6 +280,13 @@ spec:
               postStart:
                  exec:
                     command: ["/bin/sh", "-c", "sleep 2"]
+              # Removes the sock file that is used for communication between
+              # cstor-pool and cstor-pool-mgmt containers, so that, cstor-pool-mgmt
+              # of new pool pod instance will NOT communicate with cstor-pool
+              # container of terminating pod instance.
+              preStop:
+                 exec:
+                    command: ["/bin/bash", "-c", "rm /tmp/uzfs.sock"]
```
payes: Why do we share the tmp folder from the node in case of cstor pools instead of expty dir as we do in target deployment.

PR #1386, opened 2019-07-29
fix(cspc,cstor-operator): fix node selector and optimize pool expansion

sonasingh46:
** Why we need this PR? **
1. While creating a pool deployment `node selector` was not propagated to the pool deployment
    due to which there was undesired scheduling of pool pods.
    This PR fixes this.

2. While expanding the pool -- the newly added BD(s) needs to be claimed by creating BDC(s). 
    Once BDC is created for the associated BD, it can take a little to time for the BD to get `Claimed` 
    and hence an error was returned causing the reconciler to complete the expansion in the next 
    event.
    In most cases -- `k` number of events was needed to add up `k` block devices for expansion 
    that is very time-consuming.
    This PR fixes this issue by creating BDC(s) first for all the newly added BD(s) on the CSPC and 
    then go for expansion. 
    
### DIFF
--- a/cmd/maya-apiserver/cstor-operator/cspc/operations.go
+++ b/cmd/maya-apiserver/cstor-operator/cspc/operations.go
@@ -83,6 +83,13 @@ func (pc *PoolConfig) addGroupToPool(cspcPoolSpec *apis.PoolSpec, csp *apis.NewT
 			if cspcRaidGroup.Type == "" {
 				cspcRaidGroup.Type = cspcPoolSpec.PoolConfig.DefaultRaidGroupType
 			}
+
+			for _, bd := range cspcRaidGroup.BlockDevices {
+				err := pc.ClaimBD(bd.BlockDeviceName)
+				if err != nil {
+					glog.Errorf("failed to created bdc for bd %s:%s", bd.BlockDeviceName, err.Error())
+				}
+			}
 			for _, bd := range cspcRaidGroup.BlockDevices {
 				err := pc.isBDUsable(bd.BlockDeviceName)
 				if err != nil {
@@ -131,6 +138,7 @@ func (pc *PoolConfig) addBlockDeviceToGroup(group *apis.RaidGroup, csp *apis.New
 					glog.V(2).Infof("No new block devices "+
 						"added for group {%+v} on csp %s", groupOnCSP, csp.Name)
 				}
+				pc.ClaimBDList(newBDs)
 				for _, bdName := range newBDs {
 					err := pc.isBDUsable(bdName)
 					if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/maya-apiserver/cstor-operator/cspc/operations.go
```diff
@@ -83,6 +83,13 @@ func (pc *PoolConfig) addGroupToPool(cspcPoolSpec *apis.PoolSpec, csp *apis.NewT
 			if cspcRaidGroup.Type == "" {
 				cspcRaidGroup.Type = cspcPoolSpec.PoolConfig.DefaultRaidGroupType
 			}
+
```
vishnuitta: instead of this, and next for loop, how about calling `algorithmconfig.AreBDsUsable(array of BD names)` which is nothing but rename of `algorithmconfig.ClaimBDsForNode(array of BD names)`?
same applies to the next chunk of code as well

vishnuitta: Also, lets remove multi-return value from `algorithmconfig.IsClaimedBDUsable`.. lets just return error if BD is not usable.. so, at the `return` statement which is at the end of function, lets return error saying `BD is already in use`.. this avoids multiple checks in the caller

vishnuitta: lets rename `algorithmconfig.IsClaimedBDUsable` to `algorithmconfig.IsBDUsable` as the BD name that is passed might be not claimed..

cmd/maya-apiserver/cstor-operator/cspc/storagepool_create.go
```diff
@@ -109,14 +109,14 @@ func (pc *PoolConfig) GetPoolDeploySpec(csp *apis.NewTestCStorPool) (*appsv1.Dep
...
 		WithLabelsNew(getDeployLabels(csp)).
-		WithNodeSelector(csp.Spec.NodeSelector).
 		WithOwnerReferenceNew(getDeployOwnerReference(csp)).
 		WithReplicas(getReplicaCount()).
 		WithStrategyType(appsv1.RecreateDeploymentStrategyType).
 		WithSelectorMatchLabelsNew(getDeployMatchLabels()).
 		WithPodTemplateSpecBuilder(
 			pts.NewBuilder().
 				WithLabelsNew(getPodLabels(csp)).
+				WithNodeSelector(csp.Spec.NodeSelector).
```
vishnuitta: good one

cmd/maya-apiserver/cstor-operator/cspc/operations.go
sonasingh46: This may involve changes that is out of scope of this PR.
Will consider it later.

PR #1388, opened 2019-08-02
fix(bdc, cstor-operator): bdc creations in case of auto pool provisioning

mittachaitu:

This PR creates required blockdeviceclaims based on spc configurations in auto mode(after the fix).

**Before the fix:**
Before the fix we are selecting blockdevices which are in claimed and unclaimed state and later we are getting the list of blockdevice claims belongs to current spc and giving preference to blockdevices which doesn't have claims this leads to create more bdc than required count.

**After the fix:**
This fix will filter the blockdevices based which are claimed by that spc and bd's which are in unclaimed state and after getting the list of blockdeviceclaims related to spc it honors this bdc list with filtered list and give preference to node which has claims on the node. 

1. BDD for this fix is present in different PR(https://github.com/openebs/maya/pull/1394).

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #https://github.com/openebs/openebs/issues/2693

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/algorithm/nodeselect/v1alpha1/select_node.go
+++ b/pkg/algorithm/nodeselect/v1alpha1/select_node.go
@@ -153,6 +153,7 @@ func (ac *Config) selectNode(nodeBlockDeviceMap map[string]*blockDeviceList) (*n
 			Items: []string{},
 		},
 	}
+	spcObj := spcv1alpha1.SPC{Object: ac.Spc}
 	// bdCount will hold the number of block devices that will be selected from a qualified
 	// node for specific pool type
 	var bdCount int
@@ -160,7 +161,9 @@ func (ac *Config) selectNode(nodeBlockDeviceMap map[string]*blockDeviceList) (*n
 	// node for specific pool type
 	minRequiredDiskCount := spcv1alpha1.DefaultDiskCount[ac.poolType()]
 
-	if ProvisioningType(ac.Spc) == ProvisioningTypeAuto {
+	// Below snippet is required to get the node and blockdevice topology from
+	// blockdeviceclaims which were created during previous reconciliation
+	if spcObj.IsAutoProvisioning() {
 		// List all blockdeviceclaims related to spc
 		// NOTE: This list contains blockdevices on which csps' are already
 		// created filter and use it
@@ -177,22 +180,22 @@ func (ac *Config) selectNode(nodeBlockDeviceMap map[string]*blockDeviceList) (*n
 		// Form node and blockdevice topology for blockdevices which has bdc
 		claimedDevicesOnNode := customBDCList.GetBlockDeviceNamesByNode()
 		for claimedNodeName, bdNameList := range claimedDevicesOnNode {
-			qualifiedBDList := []string{}
 			filteredBlockDevices := nodeBlockDeviceMap[claimedNodeName]
 
 			//Check is this node is usable or not
 			if filteredBlockDevices != nil &&
 				util.ContainsString(filteredBlockDevices.Items, bdNameList[0]) {
-				qualifiedBDList = append(qualifiedBDList, bdNameList...)
-				if len(qualifiedBDList) < minRequiredDiskCount {
-					count := minRequiredDiskCount - len(qualifiedBDList)
-					availableBDOnNode := util.ListDiff(filteredBlockDevices.Items, qualifiedBDList)
+				// check whether bdNameList contains partially created claimed BDs
+				// i.e BD count is less than required BDs
+				if len(bdNameList) < minRequiredDiskCount {
+					count := minRequiredDiskCount - len(bdNameList)
+					availableBDOnNode := util.ListDiff(filteredBlockDevices.Items, bdNameList)
 					if len(availableBDOnNode) < count {
 						continue
 					}
-					qualifiedBDList = append(qualifiedBDList, availableBDOnNode[:count]...)
+					bdNameList = append(bdNameList, availableBDOnNode[:count]...)
 				}
-				selectedBlockDevice.BlockDevices.Items = append(selectedBlockDevice.BlockDevices.Items, qualifiedBDList...)
+				selectedBlockDevice.BlockDevices.Items = append(selectedBlockDevice.BlockDevices.Items, bdNameList...)
 				selectedBlockDevice.NodeName = claimedNodeName
 				return selectedBlockDevice, nil
 			}
@@ -255,6 +258,7 @@ func (ac *Config) poolType() string {
 func (ac *Config) getBlockDevice() (*ndmapis.BlockDeviceList, error) {
 	var bdl *blockdevice.BlockDeviceList
 	diskType := ac.Spc.Spec.Type
+	spcObj := &spcv1alpha1.SPC{Object: ac.Spc}
 	diskFilterLabel := diskFilterConstraint(diskType)
 	bdList, err := ac.BlockDeviceClient.List(metav1.ListOptions{LabelSelector: diskFilterLabel})
 	if err != nil {
@@ -268,7 +272,7 @@ func (ac *Config) getBlockDevice() (*ndmapis.BlockDeviceList, error) {
 		filterList = append(filterList, blockdevice.FilterNonSparseDevices)
 	}
 
-	if ProvisioningType(ac.Spc) == ProvisioningTypeAuto {
+	if spcObj.IsAutoProvisioning() {
 		filterList = append(filterList, blockdevice.FilterNonFSType, blockdevice.FilterNonRelesedDevices)
 	}
 
@@ -276,7 +280,9 @@ func (ac *Config) getBlockDevice() (*ndmapis.BlockDeviceList, error) {
 	if len(bdl.Items) == 0 {
 		return nil, errors.Errorf("type {%s} devices are not available to provision pools in %s mode", diskType, ProvisioningType(ac.Spc))
 	}
-	if ProvisioningType(ac.Spc) == ProvisioningTypeAuto {
+
+	// Below snippet is required when BDs are claimed but CSPs are not created
+	if spcObj.IsAutoProvisioning() {
 		newBDL, err := bdl.GetUsableBlockDevices(ac.Spc.Name, ac.Namespace)
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed to get list of usable blockdevice list")
@@ -353,12 +359,10 @@ func (ac *Config) ClaimBlockDevice(nodeBDs *nodeBlockDevice, spc *apis.StoragePo
 			if err != nil {
 				return nil, errors.Wrapf(err, "failed to build block device claim for bd {%s}", bdName)
 			}
-
 			_, err = bdcKubeclient.Create(newBDCObj.Object)
 			if err != nil {
 				return nil, errors.Wrapf(err, "failed to create block device claim for bdc {%s}", bdcName)
 			}
-			ac.VisitedNodes[hostName] = true
 			glog.Infof("successfully created block device claim {%s} for block device {%s}", bdcName, bdName)
 			// As a part of reconcilation we will create pool if all the block
 			// devices are claimed
@@ -369,6 +373,7 @@ func (ac *Config) ClaimBlockDevice(nodeBDs *nodeBlockDevice, spc *apis.StoragePo
 		claimedBD.BDName = bdObj.Name
 		nodeClaimedBDs.BlockDeviceList = append(nodeClaimedBDs.BlockDeviceList, claimedBD)
 	}
+	ac.VisitedNodes[nodeClaimedBDs.NodeName] = true
 	if pendingBDCCount != 0 {
 		return nil, errors.Errorf("pending block device claim count %d on node {%s}", pendingBDCCount, nodeClaimedBDs.NodeName)
 	}
--- a/pkg/algorithm/nodeselect/v1alpha1/select_node.go
+++ b/pkg/algorithm/nodeselect/v1alpha1/select_node.go
@@ -58,14 +58,6 @@ func (ac *Config) NodeBlockDeviceSelector() (*nodeBlockDevice, error) {
 		return nil, err
 	}
 
-	//filteredNodeBDs = nodeBlockDeviceMap
-	//if ProvisioningType(ac.Spc) == ProvisioningTypeAuto {
-	//	filteredNodeBDs, err = ac.getFilteredNodeBlockDevices(nodeBlockDeviceMap)
-	//	if err != nil {
-	//		return nil, err
-	//	}
-	//}
-
 	selectedBD, err := ac.selectNode(nodeBlockDeviceMap)
 	if err != nil {
 		return nil, err
@@ -73,83 +65,6 @@ func (ac *Config) NodeBlockDeviceSelector() (*nodeBlockDevice, error) {
 
 	return selectedBD, nil
 }
-
-//// getFilteredNodeBlockDevices returns the map of node name and block device
-//// list if no claims are present on list of block devices. If claims are present
-//// then it retunrns those block devices on which claims are created
-//func (ac *Config) getFilteredNodeBlockDevices(
-//	nodeBDs map[string]*blockDeviceList) (map[string]*blockDeviceList, error) {
-//	namespace := env.Get(env.OpenEBSNamespace)
-//	bdcKubeclient := bdc.NewKubeClient().
-//		WithNamespace(namespace)
-//	bdcList, err := bdcKubeclient.List(metav1.ListOptions{LabelSelector: string(apis.StoragePoolClaimCPK) + "=" + ac.Spc.Name})
-//	if err != nil {
-//		return nil, err
-//	}
-//	// get used node names from csp related to spc
-//	usedNodeMap, err := ac.getUsedNodeMap()
-//	if err != nil {
-//		return nil, err
-//	}
-//	//TODO: Make smaller function before removing WIP
-//
-//	// minRequiredDiskCount will hold the required number of disk that should be selected from a qualified
-//	// node for specific pool type
-//	minRequiredBDCount := blockdevice.DefaultDiskCount[ac.poolType()]
-//
-//	newNodeBDMap := make(map[string]*blockDeviceList)
-//	// form the map of node name and blockdevice list from
-//	// blockdeviceclaims created by spc
-//	for _, bdcObj := range bdcList.Items {
-//		bdcObj := bdcObj
-//		if ac.VisitedNodes[bdcObj.Spec.HostName] ||
-//			usedNodeMap[bdcObj.Spec.HostName] == 1 {
-//			continue
-//		}
-//		if newNodeBDMap[bdcObj.Spec.HostName] == nil {
-//			newNodeBDMap[bdcObj.Spec.HostName] = &blockDeviceList{
-//				Items: []string{bdcObj.Spec.BlockDeviceName},
-//			}
-//		} else {
-//			bdList := newNodeBDMap[bdcObj.Spec.HostName]
-//			bdList.Items = append(bdList.Items, bdcObj.Spec.BlockDeviceName)
-//		}
-//	}
-//	// If selectedNodeCount is zero then we can consider it is first
-//	// reconciliation and need to send
-//	selectedNodeCount := len(newNodeBDMap)
-//
-//	// Form the map of node name and blockdevice list from existing blockdevices
-//	for nodeName, bdList := range nodeBDs {
-//		// pin it
-//		nodeName := nodeName
-//		bdList := bdList
-//		if selectedNodeCount != 0 {
-//			if newNodeBDMap[nodeName] != nil {
-//				if len(newNodeBDMap[nodeName].Items) < minRequiredBDCount {
-//					// If BDC creation failed in before reconciliation we are
-//					// inserting required no blockdevices
-//					for _, bdName := range bdList.Items {
-//						if !util.ContainsString(newNodeBDMap[nodeName].Items, bdName) {
-//							bdList := newNodeBDMap[nodeName]
-//							bdList.Items = append(bdList.Items, bdName)
-//						}
-//					}
-//				}
-//			}
-//		} else {
-//			if newNodeBDMap[nodeName] == nil {
-//				newNodeBDMap[nodeName] = &blockDeviceList{
-//					Items: bdList.Items,
-//				}
-//			}
-//		}
-//	}
-//	for node, bdList := range newNodeBDMap {
-//		glog.Infof("Node-- %s BD list %#v", node, bdList)
-//	}
-//	return newNodeBDMap, nil
-//}
 
 // getUsedBlockDeviceMap gives list of disks that has already been used for pool provisioning.
 func (ac *Config) getUsedBlockDeviceMap() (map[string]int, error) {
@@ -212,7 +127,6 @@ func (ac *Config) getCandidateNode(listBlockDevice *ndmapis.BlockDeviceList) (ma
 		//If node is already visited in this reconciliation then continue with
 		//other nodes
 		if ac.VisitedNodes[nodeName] {
-			glog.Infof("-------Ignoring node: %s", nodeName)
 			continue
 		}
 		if nodeBlockDeviceMap[nodeName] == nil {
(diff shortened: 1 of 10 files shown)


pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -73,31 +73,51 @@ func (ac *Config) NodeBlockDeviceSelector() (*nodeBlockDevice, error) {
...
-				Items: bdList.Items,
+	// form the map of node name and blockdevice list from
+	// blockdeviceclaims created by spc
+	for _, bdcObj := range bdcList.Items {
+		bdcObj := bdcObj
+		if newNodeBDMap[bdcObj.Spec.HostName] == nil {
+			newNodeBDMap[bdcObj.Spec.HostName] = &blockDeviceList{
+				Items: []string{bdcObj.Spec.BlockDeviceName},
 			}
+			preferedNodeList = append(preferedNodeList, bdcObj.Spec.HostName)
```
golangcibot: SA4010: this result of append is never used, except maybe in other appends (from `staticcheck`)

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -73,31 +73,51 @@ func (ac *Config) NodeBlockDeviceSelector() (*nodeBlockDevice, error) {
...
+	// Form the map of node name and blockdevice list from existing blockdevices
+	for nodeName, bdList := range nodeBDs {
+		// pin it
+		nodeName := nodeName
+		bdList := bdList
+		if newNodeBDMap[nodeName] == nil {
+			newNodeBDMap[nodeName] = &blockDeviceList{
+				Items: bdList.Items,
 			}
+			preferedNodeList = append(preferedNodeList, nodeName)
```
golangcibot: SA4010: this result of append is never used, except maybe in other appends (from `staticcheck`)

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -73,33 +74,77 @@ func (ac *Config) NodeBlockDeviceSelector() (*nodeBlockDevice, error) {
...
+		nodeName := nodeName
+		bdList := bdList
+		if selectedNodeCount != 0 {
+			if newNodeBDMap[nodeName] != nil {
+				if len(newNodeBDMap[nodeName].Items) < minRequiredBDCount {
+					// If BDC creation failed in before reconciliation we are
+					// inserting required no blockdevices
+					for _, bdName := range bdList.Items {
+						if !util.ContainsString(newNodeBDMap[nodeName].Items, bdName) {
+							bdList := newNodeBDMap[nodeName]
```
golangcibot: shadow: declaration of "bdList" shadows declaration at line 123 (from `govet`)

pkg/util/util_test.go
```diff
@@ -55,6 +55,39 @@ func TestContainsString(t *testing.T) {
...
+		"contains string - boundary test case - similar elements but not same": {
+			listA:       []string{"hi there", "ok now"},
+			listB:       []string{},
+			expectedLen: 2,
+		},
+	}
+
+	for name, mock := range tests {
+		t.Run(name, func(t *testing.T) {
+			resultArr := ListAMinusListB(mock.listA, mock.listB)
```
golangcibot: Using the variable on range scope `mock` in function literal (from `scopelint`)

pkg/util/util_test.go
```diff
@@ -55,6 +55,39 @@ func TestContainsString(t *testing.T) {
...
+			if mock.expectedLen != len(resultArr) {
```
golangcibot: Using the variable on range scope `mock` in function literal (from `scopelint`)

pkg/util/util_test.go
```diff
@@ -55,6 +55,39 @@ func TestContainsString(t *testing.T) {
...
+				t.Fatalf("failed to test %q: expected element count '%d': actual element count '%d'", name, mock.expectedLen, len(resultArr))
```
golangcibot: Using the variable on range scope `name` in function literal (from `scopelint`)

pkg/util/util.go
```diff
@@ -132,6 +132,21 @@ func ContainsString(stringarr []string, element string) bool {
 	return false
 }
 
+// ListAMinusListB returns list of string which are in listA but not in listB
+func ListAMinusListB(listA []string, listB []string) []string {
```
sonasingh46: I would like to grab some attention on the function name here. ( Optional )
One suggestion : 
`ListDiff`

pkg/blockdevice/v1alpha1/blockdevice.go
```diff
@@ -211,6 +212,34 @@ func (bdl *BlockDeviceList) Filter(predicateKeys ...string) *BlockDeviceList {
 	return filteredBlockDeviceList
 }
 
+// GetUsableBlockDevices returns custom blockdevice list which can be used
```
sonasingh46: What is a custom block device list ?

pkg/blockdevice/v1alpha1/blockdevice.go
```diff
@@ -211,6 +212,34 @@ func (bdl *BlockDeviceList) Filter(predicateKeys ...string) *BlockDeviceList {
...
+func (bdl *BlockDeviceList) GetUsableBlockDevices(spcName, namespace string) (*BlockDeviceList, error) {
+	// Initialize filtered block device list
+	if bdl == nil {
+		return nil, errors.Errorf("failed to get usable blockdevices: blockdevicelist is nil")
```
sonasingh46: 1.Why there is a chance of this function called on `nil` pointer receivers?

pkg/blockdevice/v1alpha1/blockdevice.go
```diff
@@ -211,6 +212,34 @@ func (bdl *BlockDeviceList) Filter(predicateKeys ...string) *BlockDeviceList {
...
+		BlockDeviceList: &ndm.BlockDeviceList{},
+		errs:            nil,
+	}
+	bdcClient := bdc_v1alpha1.NewKubeClient().WithNamespace(namespace)
+	for _, bdObj := range bdl.Items {
+		if bdObj.Status.ClaimState == ndm.BlockDeviceClaimed {
+			bdcName := bdObj.Spec.ClaimRef.Name
+			bdcObj, err := bdcClient.Get(bdcName, metav1.GetOptions{})
+			if err != nil {
+				return nil, err
```
sonasingh46: It is better to add a little context to the error here .

pkg/blockdevice/v1alpha1/blockdevice.go
```diff
@@ -211,6 +212,34 @@ func (bdl *BlockDeviceList) Filter(predicateKeys ...string) *BlockDeviceList {
...
+			}
+			if bdcObj.Labels[string(apis.StoragePoolClaimCPK)] == spcName {
+				filteredBlockDeviceList.Items = append(filteredBlockDeviceList.Items, bdObj)
+			}
+		} else if bdObj.Status.ClaimState == ndm.BlockDeviceUnclaimed {
+			filteredBlockDeviceList.Items = append(filteredBlockDeviceList.Items, bdObj)
+		}
```
sonasingh46: The repetition in `else if`  block can be avoided.

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -263,6 +275,13 @@ func (ac *Config) getBlockDevice() (*ndmapis.BlockDeviceList, error) {
 	if len(bdl.Items) == 0 {
 		return nil, errors.Errorf("type {%s} devices are not available to provision pools in %s mode", diskType, ProvisioningType(ac.Spc))
 	}
+	if ProvisioningType(ac.Spc) == ProvisioningTypeAuto {
+		newBDL, err := bdl.GetUsableBlockDevices(ac.Spc.Name, ac.Namespace)
+		if err != nil {
+			return nil, err
```
sonasingh46: A little context to the error.

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -186,6 +159,45 @@ func (ac *Config) selectNode(nodeBlockDeviceMap map[string]*blockDeviceList) *no
...
+			WithNamespace(ac.Namespace).
+			List(metav1.ListOptions{LabelSelector: string(apis.StoragePoolClaimCPK) + "=" + ac.Spc.Name})
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to select node and blockdevices")
+		}
+		customBDCList := bdc.BlockDeviceClaimList{
+			ObjectList: bdcList,
+		}
+		// Form node and blockdevice topology for blockdevices which has bdc
+		claimedNodes := customBDCList.GetBlockDeviceNamesByNode()
```
sonasingh46: `claimedNodes` can confuse.

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -186,6 +159,45 @@ func (ac *Config) selectNode(nodeBlockDeviceMap map[string]*blockDeviceList) *no
...
+		// List all blockdeviceclaims related to spc
+		// NOTE: This list contains blockdevices on which csps' are already
+		// created filter and use it
+		bdcList, err := bdc.NewKubeClient().
+			WithNamespace(ac.Namespace).
+			List(metav1.ListOptions{LabelSelector: string(apis.StoragePoolClaimCPK) + "=" + ac.Spc.Name})
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to select node and blockdevices")
+		}
+		customBDCList := bdc.BlockDeviceClaimList{
```
sonasingh46: What is `customBDCList` ? Either we can make the var name more intuitive or put some comment.

pkg/util/util.go
mittachaitu: I will update it

pkg/blockdevice/v1alpha1/blockdevice.go
mittachaitu: Custom block device list is wrapper over core api object list

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: I will update with comments

pkg/blockdevice/v1alpha1/blockdevice.go
mittachaitu: Not sure why I added this check. I will think about it.

pkg/blockdevice/v1alpha1/blockdevice.go
mittachaitu: If we avoid it how can we get unclaimed blockdevices?

pkg/util/util.go
mittachaitu: Done

pkg/blockdevice/v1alpha1/blockdevice.go
mittachaitu: Done

pkg/blockdevice/v1alpha1/blockdevice.go
mittachaitu: removed the check

pkg/blockdevice/v1alpha1/blockdevice.go
mittachaitu: Updated

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: Updated

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: updated

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: Done

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -263,6 +276,13 @@ func (ac *Config) getBlockDevice() (*ndmapis.BlockDeviceList, error) {
 	if len(bdl.Items) == 0 {
 		return nil, errors.Errorf("type {%s} devices are not available to provision pools in %s mode", diskType, ProvisioningType(ac.Spc))
 	}
+	if ProvisioningType(ac.Spc) == ProvisioningTypeAuto {
+		newBDL, err := bdl.GetUsableBlockDevices(ac.Spc.Name, ac.Namespace)
```
vishnuitta: why we need this? in what cases, this helps us

mittachaitu: When user created `n spc` at a time then controller will process one by one. Let us assume that `spc1` is processed and cstor-operator created claim when cstor-operator picked up second `spc2` to create pools then we shouldn't include the `bd` which is claimed for `spc-1`.

vishnuitta: if claims are not fulfilled, still, there are chances of picking a BD for 2 SPCs right?

mittachaitu: not fulfilled means not yet bounded? In this case Yes there are chances of getting that blockdevice but before using to create a pool we have strict check it should be claimed by spc which is processing.

vishnuitta: yes.. I meant - not fulfilled as 'not yet bound'
and I also saw other code where checks are there that BDs are claimed by SPC before creating pool
If CSP is already created, then, those BDs are already filtered out.
I'm trying to understand the need of this code.
This code will NOT cover the case that I mentioned above.
And, this code might cover the case where BDs are claimed but CSPs are not created yet. Is my understanding right?

mittachaitu: > If CSP is already created, then, those BDs are already filtered out.

No we are doing this filter in `getCandidateNode`

> And, this code might cover the case where BDs are claimed but CSPs are not created yet. Is my understanding right?

Yes this will cover above case.

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -186,6 +159,46 @@ func (ac *Config) selectNode(nodeBlockDeviceMap map[string]*blockDeviceList) *no
...
+			return nil, errors.Wrapf(err, "failed to select node and blockdevices")
+		}
+		// customBDCList is a wrapper over blockdeviceclaim list
+		customBDCList := bdc.BlockDeviceClaimList{
+			ObjectList: bdcList,
+		}
+		// Form node and blockdevice topology for blockdevices which has bdc
+		claimedDevicesOnNode := customBDCList.GetBlockDeviceNamesByNode()
+		for claimedNodeName, bdNameList := range claimedDevicesOnNode {
+			qualifiedBDList := []string{}
```
vishnuitta: why we need this variable? can't we directly use bdNameList

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -263,6 +276,13 @@ func (ac *Config) getBlockDevice() (*ndmapis.BlockDeviceList, error) {
 	if len(bdl.Items) == 0 {
 		return nil, errors.Errorf("type {%s} devices are not available to provision pools in %s mode", diskType, ProvisioningType(ac.Spc))
 	}
+	if ProvisioningType(ac.Spc) == ProvisioningTypeAuto {
```
vishnuitta: adding a comment on why we need this code would be good

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -186,6 +159,46 @@ func (ac *Config) selectNode(nodeBlockDeviceMap map[string]*blockDeviceList) *no
 	// minRequiredDiskCount will hold the required number of disk that should be selected from a qualified
 	// node for specific pool type
 	minRequiredDiskCount := spcv1alpha1.DefaultDiskCount[ac.poolType()]
+
+	if ProvisioningType(ac.Spc) == ProvisioningTypeAuto {
```
vishnuitta: here as well, a comment on why this 'if' check required would be good

vishnuitta: can this be made as function?

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -338,6 +358,7 @@ func (ac *Config) ClaimBlockDevice(nodeBDs *nodeBlockDevice, spc *apis.StoragePo
 			if err != nil {
 				return nil, errors.Wrapf(err, "failed to create block device claim for bdc {%s}", bdcName)
 			}
+			ac.VisitedNodes[hostName] = true
```
vishnuitta: you better set this when selectNode picked it

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -186,6 +159,46 @@ func (ac *Config) selectNode(nodeBlockDeviceMap map[string]*blockDeviceList) *no
...
+				qualifiedBDList = append(qualifiedBDList, bdNameList...)
+				if len(qualifiedBDList) < minRequiredDiskCount {
+					count := minRequiredDiskCount - len(qualifiedBDList)
+					availableBDOnNode := util.ListDiff(filteredBlockDevices.Items, qualifiedBDList)
+					if len(availableBDOnNode) < count {
+						continue
+					}
+					qualifiedBDList = append(qualifiedBDList, availableBDOnNode[:count]...)
+				}
+				selectedBlockDevice.BlockDevices.Items = append(selectedBlockDevice.BlockDevices.Items, qualifiedBDList...)
```
vishnuitta: I think - this causes problem. Here, we are selecting disks which can be out of filterBlockDevices. Can you please look into this?
this is similar to the issue that seen previously.

vishnuitta: if I'm not wrong, this need to be intersection of filteredBlockDevices and bdNameList

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: Yes we can reuse it.

mittachaitu: I will update it

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: Yes, I will update it

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: No It won't because we are checking whether it can be usable or not by looking into filteredBlockdeviceList.
Link of check
https://github.com/openebs/maya/blob/d2ea21dc4433b51ee6d559bb96c251101316006b/pkg/algorithm/nodeselect/v1alpha1/select_node.go#L185

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: Basically it is useful only in partial bdc creation case instead of modifying `bdNameList`. Anyway I will update it.

pkg/algorithm/nodeselect/v1alpha1/select_node.go
vishnuitta: we are checking for only first disk

mittachaitu: I think this is good enough because we are forming node and blockdevice topology checking on one BD is good to enough to exclude it from creating pool again.

vishnuitta:
if BDD for this fix is already merged, then, that BDD should be failing without this fix.. is that the case?

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: >  If I'm not wrong, this need to be intersection of filteredBlockDevices and bdNameList

In partial claims creations case intersection won't helpful we will get bd count less than required bd count

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -150,42 +115,94 @@ func (ac *Config) getCandidateNode(listBlockDevice *ndmapis.BlockDeviceList) (ma
...
+				// i.e BD count is less than required BDs
+				if len(qualifiedBDList) < minRequiredDiskCount {
+					count := minRequiredDiskCount - len(qualifiedBDList)
+					availableBDOnNode := util.ListDiff(filteredBlockDevices.Items, qualifiedBDList)
+					if len(availableBDOnNode) < count {
+						continue
+					}
+					qualifiedBDList = append(qualifiedBDList, availableBDOnNode[:count]...)
+				}
+				selectedBlockDevice.BlockDevices.Items = append(selectedBlockDevice.BlockDevices.Items, qualifiedBDList...)
```
vishnuitta: should we make this as
`selectedBlockDevice.BlockDevices.Items = append(selectedBlockDevice.BlockDevices.Items, qualifiedBDList[:minRequiredDiskCount]...)`

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -348,6 +374,7 @@ func (ac *Config) ClaimBlockDevice(nodeBDs *nodeBlockDevice, spc *apis.StoragePo
 		claimedBD.BDName = bdObj.Name
 		nodeClaimedBDs.BlockDeviceList = append(nodeClaimedBDs.BlockDeviceList, claimedBD)
 	}
+	ac.VisitedNodes[nodeClaimedBDs.NodeName] = true
```
vishnuitta: we thought of moving this to selectNode

pkg/util/util.go
```diff
@@ -132,6 +132,36 @@ func ContainsString(stringarr []string, element string) bool {
 	return false
 }
 
+// ListDiff returns list of string which are in listA but not in listB
```
vishnuitta: these two functions are good.. but, think of `smart coding` here

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: Updated

pkg/algorithm/nodeselect/v1alpha1/select_node.go
mittachaitu: Updated

pkg/util/util.go
mittachaitu: Added as `TODO` will be done in next PRs

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -150,42 +115,105 @@ func (ac *Config) getCandidateNode(listBlockDevice *ndmapis.BlockDeviceList) (ma
...
+			ObjectList: bdcList,
+		}
+		// Form node and blockdevice topology for blockdevices which has bdc
+		claimedDevicesOnNode := customBDCList.GetBlockDeviceNamesByNode()
+		for claimedNodeName, bdNameList := range claimedDevicesOnNode {
+			filteredBlockDevices := nodeBlockDeviceMap[claimedNodeName]
+
+			//Check is this node is usable or not
+			if filteredBlockDevices != nil {
+				qualifiedBDList := []string{}
```
golangcibot: SA4006: this value of `qualifiedBDList` is never used (from `staticcheck`)

pkg/algorithm/nodeselect/v1alpha1/select_node.go
```diff
@@ -161,13 +163,19 @@ func (ac *Config) selectNode(nodeBlockDeviceMap map[string]*blockDeviceList) (*n
 	// node for specific pool type
 	minRequiredDiskCount := spcv1alpha1.DefaultDiskCount[ac.poolType()]
 
+	defer func() {
+		if err == nil {
+			ac.VisitedNodes[selectedBlockDevice.NodeName] = true
```
vishnuitta: will this work even if NodeName is empty ""?
This can happen if nothing from nodeBlockDeviceMap is selected.

mittachaitu: > this work even if NodeName is empty ""?

Yes it will work but removed `defer()`

PR #1416, opened 2019-08-29
fix(cspi): update status parsing and get devicePath ignoring "/dev"

prateekpandey14:

fix the status parsing for cstorpoolinstance resource.
```sh
kubectl get cspi -n openebs
NAME               HOSTNAME    ALLOCATED   FREE    CAPACITY   STATUS   AGE
cspc-sparse-2kht   127.0.0.1   281K        9.94G   9.94G      ONLINE   13m

```

### DIFF
--- a/cmd/cstor-pool-mgmt/pool/pool.go
+++ b/cmd/cstor-pool-mgmt/pool/pool.go
@@ -172,7 +172,7 @@ func GetDevPathIfNotSlashDev(devID string) string {
 	}
 	lastindex := strings.LastIndexByte(devID, '/')
 	if lastindex == -1 {
-		lastindex = len(devID)
+		return ""
 	}
 	devidbytes := []rune(devID)
 	return string(devidbytes[0:lastindex])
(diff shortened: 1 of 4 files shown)


cmd/cstor-pool-mgmt/pool/pool.go
```diff
@@ -160,23 +160,31 @@ func importPoolBuilder(cStorPool *apis.CStorPool, importOptions *ImportOptions)
...
+		return ""
+	}
+
+	if strings.HasPrefix(devID, "/dev") {
+		return ""
+	}
+	lastindex := strings.LastIndexByte(devID, '/')
 	if lastindex == -1 {
-		lastindex = len(devid)
+		lastindex = len(devID)
```
mynktl: Without `/` means only disk name, like `sda/b/c`. In that case, it should return `empty`. Though this case is not possible, we should return empty here IMO.

prateekpandey14: Done

PR #1441, opened 2019-09-11
chore(log,client-go): replace golang/glog with k8s.io/klog and bump to client-go 1.14.3

prateekpandey14:
- We need this change as part of stablilty and secuirity fixes has been done in kubernetes 1.14+ 
- As 1.16 relreased, minor releases occur approximately every 3 months, so each minor release 
   branch is maintained for approximately 9 months, k8s 1.12 branch is not maintenained further.
- We need some stabilty fixes which are part of 1.14+ release like leader-election API , admission 
 controller etc.
- All the cloud provider started supporting k8s 1.14 versions.
- With Openebs CSI driver we are already using k8s 1.14.3 version and it will be good to have same 
  k8s version.
- k8s fork the `golang/glog` and maintaining as `k8s.io/klog` and replaced glog in k8s version v1.14+ onwards, to upgrade in k8s version 1.14 or more we have to use either klog or any other logging mechanism other than glog.
 
Commits changes are below:
- Updates the kubernetes deps to 1.14.3 version
- Update the client-go version to 1.14.3 version
- Update external-storage-lib to v3.1.0 version with respect to 1.14.3 k8s deps
- Replace golang/glog --> k8s.io/klog


cmd/admission-server/main.go
```diff
@@ -45,33 +45,36 @@ func main() {
 	flag.IntVar(&parameters.Port, "port", 443, "Webhook server port.")
 	flag.StringVar(&parameters.CertFile, "tlsCertFile", "/etc/webhook/certs/cert.pem", "File containing the x509 Certificate for HTTPS.")
 	flag.StringVar(&parameters.KeyFile, "tlsKeyFile", "/etc/webhook/certs/key.pem", "File containing the x509 private key to --tlsCertFile.")
+
+	klog.InitFlags(nil)
+	flag.Set("logtostderr", "true")
```
golangcibot: Error return value of `flag.Set` is not checked (from `errcheck`)

PR #1445, opened 2019-09-17
chrore(cspc): enable cspc-operator to take lease for the lifetime on cspc

sonasingh46:
This PR will enable highly available lease for cspc-operator pod.
Once a cspc-operator pod acquires a lease on CSPC object, only that pod will be able to process and no other pod will be able to process the CSPC with lease.
If in case, the leader ( leader is the pod who has acquired the lease on the CSPC ) has died another pod can acquire CSPC with lease for processing it.

### DIFF
--- a/cmd/cspc-operator/app/cspc_lease.go
+++ b/cmd/cspc-operator/app/cspc_lease.go
@@ -82,6 +82,10 @@ func (sl *Lease) Hold() error {
 			return err
 		}
 	}
+	// If the pod which has already acquired lease and want again to acquire,grant it.
+	if leaseValueObj.Holder == sl.getPodName() {
+		return nil
+	}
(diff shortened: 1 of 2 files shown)


cmd/cspc-operator/app/cspc_lease.go
```diff
@@ -82,6 +82,10 @@ func (sl *Lease) Hold() error {
 			return err
 		}
 	}
+	// If the pod which has already acquired lease and want again to acquire,grant it.
+	if leaseValueObj.Holder == env.Get(env.OpenEBSNamespace)+"/"+env.Get(env.CSPCOperatorPodName) {
```
vishnuitta: can we use sl.getPodName()?

sonasingh46: done

PR #1458, opened 2019-09-26
feat(cspc_provisioning_bdd) : add positive bdd test cases for cspc provisioning

sonasingh46:
This PR adds positive test cases for CSPC provisioning.

### DIFF
--- a/tests/cstor/cspc/positive/provisioning/cspc_provisioning_utils_test.go
+++ b/tests/cstor/cspc/positive/provisioning/cspc_provisioning_utils_test.go
@@ -47,7 +47,7 @@ func createCSPCObject(blockDeviceCount int, poolType string) {
 			WithRaidGroupBuilder(
 				cspcrg_v1alpha1.NewBuilder().
 					// TODO : PAss the entire label -- kubernetes.io/hostname
-					WithCSPCBlockDeviceList(ops.GetCSPCBDListForNode(NodeList.Items[0].Name, blockDeviceCount)).
+					WithCSPCBlockDeviceList(ops.GetCSPCBDListForNode(&NodeList.Items[0], blockDeviceCount)).
 					WithType(poolType),
 			),
 		).
@@ -55,7 +55,7 @@ func createCSPCObject(blockDeviceCount int, poolType string) {
 			WithNodeSelector(NodeList.Items[1].Labels).
 			WithRaidGroupBuilder(
 				cspcrg_v1alpha1.NewBuilder().
-					WithCSPCBlockDeviceList(ops.GetCSPCBDListForNode(NodeList.Items[1].Name, blockDeviceCount)).
+					WithCSPCBlockDeviceList(ops.GetCSPCBDListForNode(&NodeList.Items[1], blockDeviceCount)).
 					WithType(poolType),
 			),
 		).
@@ -63,7 +63,7 @@ func createCSPCObject(blockDeviceCount int, poolType string) {
 			WithNodeSelector(NodeList.Items[2].Labels).
 			WithRaidGroupBuilder(
 				cspcrg_v1alpha1.NewBuilder().
-					WithCSPCBlockDeviceList(ops.GetCSPCBDListForNode(NodeList.Items[2].Name, blockDeviceCount)).
+					WithCSPCBlockDeviceList(ops.GetCSPCBDListForNode(&NodeList.Items[2], blockDeviceCount)).
 					WithType(poolType),
 			),
 		).
--- a/tests/cstor/cspc/positive/provisioning/cspc_provisioning_utils_test.go
+++ b/tests/cstor/cspc/positive/provisioning/cspc_provisioning_utils_test.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2019 The OpenEBS Authors
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package provisioning
+
+import (
+	. "github.com/onsi/gomega"
+	apis "github.com/openebs/maya/pkg/apis/openebs.io/v1alpha1"
+	cspc_v1alpha1 "github.com/openebs/maya/pkg/cstor/poolcluster/v1alpha1"
+	cspcspecs_v1alpha1 "github.com/openebs/maya/pkg/cstor/poolcluster/v1alpha1/cstorpoolspecs"
+	cspcrg_v1alpha1 "github.com/openebs/maya/pkg/cstor/poolcluster/v1alpha1/raidgroups"
+)
+
+func createCSPCObjectForStripe()  {
+	createCSPCObject(1,"stripe")
+}
+
+func createCSPCObjectForMirror()  {
+	createCSPCObject(2,"mirror")
+}
+
+func createCSPCObjectForRaidz()  {
+	createCSPCObject(3,"raidz")
+}
+
+func createCSPCObjectForRaidz2()  {
+	createCSPCObject(6,"raidz2")
+}
+
+func createCSPCObject(blockDeviceCount int,poolType string)  {
+	var err error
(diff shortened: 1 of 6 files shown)


tests/cstor/cspc/positive/provisioning/provisioning_test.go
```diff
@@ -0,0 +1,97 @@
...
+	provisioningAndReconciliationTest(createCSPCObjectForMirror)
+})
+var _ = Describe("[CSPC] CSTOR RAIDZ POOL PROVISIONING AND RECONCILIATION ", func() {
+	provisioningAndReconciliationTest(createCSPCObjectForRaidz)
+})
+var _ = Describe("[CSPC] CSTOR RAIDZ2 POOL PROVISIONING AND RECONCILIATION ", func() {
+	provisioningAndReconciliationTest(createCSPCObjectForRaidz2)
+})
+
+func provisioningAndReconciliationTest(createCSPCObject func())  {
```
vishnuitta: Good one @sonasingh46 :+1: 
It would be good idea if we can create issues for BDD testcases which need to be covered.
Please add comments, so that new contributors would be able to add testcases

tests/cstor/cspc/positive/provisioning/provisioning_test.go
```diff
@@ -0,0 +1,97 @@
...
+			// We expect 3 cstorPool objects.
+			cspCount := ops.GetHealthyCSPICount(cspcObj.Name, 3)
+			Expect(cspCount).To(Equal(3))
+		})
+	})
+
+	////Test Case #2 : Dependent on above test case #1 . | TestType : Reconciliation
+	When("3 CSPIs Is Deleted", func() {
+		It("3 New CSPIs Should Come Up Again", func() {
+			ops.DeleteCSPI(cspcObj.Name, 1)
```
shubham14bajpai: Can we have `ops.DeleteCSPI(cspcObj.Name, 1)` instead to delete 3 CSPI.

tests/cstor/cspc/positive/provisioning/provisioning_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	When("The CSPC Finalizer Is Removed From CSPC", func() {
+		It("The Finalizer Should Come Back As Part Of Reconcilation", func() {
+			err := Cspc.RemoveFinalizer(cspc_v1alpha1.CSPCFinalizer)
+			Expect(err).To(BeNil())
+			Expect(ops.IsCSPCFinalizerExistsOnCSPC(cspcObj.Name, cspc_v1alpha1.CSPCFinalizer)).To(BeTrue())
+		})
+	})
+	// TODO : Add test case for pool import
+	When("1 CSPI Is Deleted", func() {
```
shubham14bajpai: Instead of 3 tests can we have a for loop and take poolCount as a flag ?
```golang
for i := 1; i <= poolCount; i++ {
		When(strconv.Itoa(i)+" CSPIs Is Deleted", func() {
		It(strconv.Itoa(i)+" New CSPIs Should Come Up Again", func() {
			ops.DeleteCSPI(cspcObj.Name, i)
			// We expect i cstorPool objects.
			cspCount := ops.GetHealthyCSPICount(cspcObj.Name, poolCount)
			Expect(cspCount).To(Equal(poolCount))
		})
	})
	}
```
This will allow to run the test for different number of pools.
```

File: tests/operations.go [hunk 4]
```diff
@@ -870,3 +947,21 @@ func (ops *Operations) GetBDCCount(lSelector, namespace string) int {
 	Expect(err).ShouldNot(HaveOccurred())
 	return len(bdcList.Items)
 }
+
+// GetCSPCBDListForNode returns unclaimed block devices that can be used.
+func (ops *Operations)GetCSPCBDListForNode(nodeName string, blockDeviceCount int) []*apis.CStorPoolClusterBlockDevice {
+	bdList,err:=ops.BDClient.WithNamespace(ops.NameSpace).List(metav1.ListOptions{LabelSelector:"kubernetes.io/hostname="+nodeName})
```
mittachaitu: This line has some cluster dependency(some platforms have nodeName and hostName are different ex: openshift new versions)

tests/cstor/cspc/positive/provisioning/cspc_provisioning_utils_test.go
```diff
@@ -0,0 +1,90 @@
...
+	Expect(err).ShouldNot(HaveOccurred())
+	cspcObj,err = ops.CSPCClient.WithNamespace(ops.NameSpace).Create(cspcObj)
+	Expect(err).To(BeNil())
+
+	Cspc,err = cspc_v1alpha1.BuilderForAPIObject(cspcObj).Build()
+	Expect(err).To(BeNil())
+}
+
+func verifyDesiredCSPICount()  {
+	cspiCount := ops.GetHealthyCSPICount(cspcObj.Name, 3)
```
mittachaitu: Remove hardcoded values let us take input from cli

tests/cstor/cspc/positive/provisioning/provisioning_test.go
```diff
@@ -0,0 +1,97 @@
...
+})
+var _ = Describe("[CSPC] CSTOR RAIDZ2 POOL PROVISIONING AND RECONCILIATION ", func() {
+	provisioningAndReconciliationTest(createCSPCObjectForRaidz2)
+})
+
+func provisioningAndReconciliationTest(createCSPCObject func())  {
+	When("A CSPC Is Created", func() {
+		It("cStor Pools Should be Provisioned ", func() {
+
+			By("Preparing A CSPC Object, No Error Should Occur", createCSPCObject)
```
mittachaitu: We should also verify blockdevice claimed for this cspc creation. It is kind of enhancement

tests/cstor/cspc/positive/provisioning/provisioning_test.go
```diff
@@ -0,0 +1,97 @@
...
+	When("Cleaning up cspc", func() {
+		It("should delete the cspc", func() {
+			err := ops.CSPCClient.Delete(cspcObj.Name, &metav1.DeleteOptions{})
+			Expect(err).To(BeNil())
+			bdcCount := ops.GetBDCCountEventually(
+				metav1.ListOptions{
+					LabelSelector: string(apis.CStorPoolClusterCPK) + "=" + cspcObj.Name},
+				0, string(artifacts.OpenebsNamespace))
+			Expect(bdcCount).To(BeZero())
+			Expect(ops.IsSPCNotExists(cspcObj.Name)).To(BeTrue())
```
mittachaitu: Why we need this check in cspc provisioning?

tests/cstor/cspc/positive/provisioning/provisioning_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// TODO : Improve this cleanup BDD
+	When("Cleaning up cspc", func() {
+		It("should delete the cspc", func() {
+			err := ops.CSPCClient.Delete(cspcObj.Name, &metav1.DeleteOptions{})
+			Expect(err).To(BeNil())
+			bdcCount := ops.GetBDCCountEventually(
+				metav1.ListOptions{
+					LabelSelector: string(apis.CStorPoolClusterCPK) + "=" + cspcObj.Name},
+				0, string(artifacts.OpenebsNamespace))
```
mittachaitu: can we use namespace which was declared in suite file?

tests/cstor/cspc/positive/provisioning/provisioning_test.go
sonasingh46: This should be fixed.

tests/cstor/cspc/positive/provisioning/provisioning_test.go
sonasingh46: Wll fix it.

tests/cstor/cspc/positive/provisioning/provisioning_test.go
sonasingh46: Yeah -- We will put it in some later PR as part of other test cases.

tests/cstor/cspc/positive/provisioning/cspc_provisioning_utils_test.go
sonasingh46: As discussed, we need to create some way around CSPC object creation so as to run on single-node/ ( or n-node ) cluster, after that this value can be passed via flag. This will be done in a separate PR.

sonasingh46:
Also -- One thing that came out from this PR is fsType filter for block devices after having. As of now CSPC itself is not having fsType filters -- once it is added we can add this to bdd as well.

PR #1461, opened 2019-10-01
Fix shell quoting in livenessprobe command in csp yaml

singhmeghna79:
Signed-off-by: Meghna Singh <meghna.singh@mayadata.io>

With this PR the command in livenessProbe in cstorpool yaml will set the value of  `io.openebs:livenesstimestamp` property to current timestamp, whose value earlier was set to $date due to some quoting issue in the command. This PR fixes it.


vishnuitta:
upgrade need to be taken care? cc: @shubham14bajpai

PR #1448, opened 2019-09-19
refact(cv,cvr): added code to populate versiondetails and upgrade steps

shubham14bajpai:
Signed-off-by: shubham <shubham.bajpai@mayadata.io>

**What this PR does / why we need it**:
This PR adds code to handle the new spec versiondetails and also the spec changes required for features to come in this release.
The upgrade steps cannot be performed from outside the controller as the reconcile function will still have the old schema. This steps can only be performed after the image is updated for corresponding resource.

The PR also contain the schema changes for cv and cvr as the feature PR is not yet checked in.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Labelled this PR & related issue with `documentation` tag
- [ ] PR messages has document related information
- [ ] Labelled this PR & related issue with `breaking-changes` tag
- [ ] PR messages has breaking changes related information
- [ ] Labelled this PR & related issue with `requires-upgrade` tag
- [ ] PR messages has upgrade related information
- [ ] Commit has unit tests
- [ ] Commit has integration tests

PR #1465, opened 2019-10-04
refact(cv,cvr,cspc,cspi): populate version details for csi based provisioning

prateekpandey14:

changes uses new spec versiondetails and set the current and
desired versions for cstorvolume,cstorvolumereplica, cspc and cstorpool
instances resources in case of csi based volume and pool provisioning.

more info #1448

### DIFF
--- a/cmd/cspc-operator/app/handler.go
+++ b/cmd/cspc-operator/app/handler.go
@@ -18,12 +18,15 @@ package app
 
 import (
 	"fmt"
+
 	bdc "github.com/openebs/maya/pkg/blockdeviceclaim/v1alpha1"
 	apiscspc "github.com/openebs/maya/pkg/cstor/poolcluster/v1alpha1"
+	"github.com/openebs/maya/pkg/version"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	apiscsp "github.com/openebs/maya/pkg/cstor/poolinstance/v1alpha3"
 	"time"
+
+	apiscsp "github.com/openebs/maya/pkg/cstor/poolinstance/v1alpha3"
 
 	nodeselect "github.com/openebs/maya/pkg/algorithm/nodeselect/v1alpha2"
 	apis "github.com/openebs/maya/pkg/apis/openebs.io/v1alpha1"
@@ -117,6 +120,12 @@ func (c *Controller) syncCSPC(cspcGot *apis.CStorPoolCluster) error {
 		c.recorder.Event(cspcGot, corev1.EventTypeWarning, "Getting Namespace", message)
 		klog.Errorf("Could not sync CSPC {%s}: got empty namespace for openebs from env variable", cspcGot.Name)
 		return nil
+	}
+
+	_, err := c.populateVersion(cspcGot)
+	if err != nil {
+		klog.Errorf("failed to add versionDetails to CSPC %s:%s", cspcGot.Name, err.Error())
+		return err
 	}
 
 	pc, err := c.NewPoolConfig(cspcGot, openebsNameSpace)
@@ -354,4 +363,59 @@ func (pc *PoolConfig) removeSPCFinalizerOnAssociatedBDC() error {
 	}
 
 	return nil
+}
+
+// populateVersion assigns VersionDetails for old cspc object and newly created
+// cspc
+func (c *Controller) populateVersion(cspc *apis.CStorPoolCluster) (*apis.CStorPoolCluster, error) {
+	if cspc.VersionDetails.Current == "" {
+		var err error
+		var v string
+		var obj *apis.CStorPoolCluster
+		v, err = c.EstimateCSPCVersion(cspc)
+		if err != nil {
+			return nil, err
+		}
+		cspc.VersionDetails.Current = v
+		// For newly created spc Desired field will also be empty.
+		cspc.VersionDetails.Desired = v
+		cspc.VersionDetails.DependentsUpgraded = true
+		obj, err = c.clientset.OpenebsV1alpha1().
+			CStorPoolClusters(env.Get(env.OpenEBSNamespace)).
+			Update(cspc)
+
+		if err != nil {
+			return nil, errors.Wrapf(
+				err,
+				"failed to update spc %s while adding versiondetails",
+				cspc.Name,
+			)
+		}
+		klog.Infof("Version %s added on spc %s", v, cspc.Name)
+		return obj, nil
+	}
+	return cspc, nil
+}
+
+// EstimateCSPCVersion returns the cspi version if any cspi is present for the cspc or
+// returns the maya version as the new cspi created will be of maya version
+func (c *Controller) EstimateCSPCVersion(cspc *apis.CStorPoolCluster) (string, error) {
+
+	cspiList, err := c.clientset.OpenebsV1alpha1().
+		CStorPoolInstances(env.Get(env.OpenEBSNamespace)).
+		List(
+			metav1.ListOptions{
+				LabelSelector: string(apis.CStorPoolClusterCPK) + "=" + cspc.Name,
+			})
+	if err != nil {
+		return "", errors.Wrapf(
+			err,
+			"failed to get the cstorpool instance list related to cspc : %s",
+			cspc.Name,
+		)
+	}
+	if len(cspiList.Items) == 0 {
+		return version.Current(), nil
+	}
+	return cspiList.Items[0].Labels[string(apis.OpenEBSVersionKey)], nil
 }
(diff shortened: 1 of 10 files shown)


cmd/cspc-operator/app/handler.go
```diff
@@ -119,6 +122,12 @@ func (c *Controller) syncCSPC(cspcGot *apis.CStorPoolCluster) error {
 		return nil
 	}
 
+	_, err := c.populateVersion(cspcGot)
```
vishnuitta: you better take the object returned here

pkg/cstorvolumeclaim/v1alpha1/build.go
```diff
@@ -303,15 +303,27 @@ func (b *Builder) WithCapacity(capacity string) *Builder {
 	return b.WithCapacityQty(resCapacity)
 }
 
-// WithCapacityQty sets Capacity of CStorVOlumeClaim
+// WithCapacityQty sets Capacity of CStorVolumeClaim
 func (b *Builder) WithCapacityQty(resCapacity resource.Quantity) *Builder {
-	resourceList := metav1.ResourceList{
```
vishnuitta: there were discussions around this.. can this get approved from @AmitKumarDas or @mittachaitu ?

cmd/cstorvolumeclaim/cstorvolumeclaim.go
```diff
@@ -437,3 +442,27 @@ func randomizePoolList(list *apis.CStorPoolInstanceList) *apis.CStorPoolInstance
...
+// CVCWithVersionDetails build cvc resource with required version
+// details and volume references
+func CVCWithVersionDetails(
+	claim *apis.CStorVolumeClaim,
+	volumeRef *corev1.ObjectReference,
+) (*apis.CStorVolumeClaim, error) {
+
+	cvcObj, err := cvc.BuildFrom(claim).
+		WithClaimRef(volumeRef).
+		WithStatusCapacity(claim.Spec.Capacity).
```
vishnuitta: we are filling capacity as well? aren't we planned it as part of resize code path getting executed during volume creation?

cmd/cstorvolumeclaim/cstorvolumeclaim.go
```diff
@@ -367,6 +370,8 @@ func createCVR(
 			WithFinalizers(getCVRFinalizer()).
 			WithTargetIP(service.Spec.ClusterIP).
 			WithCapacity(volume.Spec.Capacity.String()).
+			WithNewVersion(version.GetVersion()).
```
vishnuitta: is setting newVersion handled for replicas created from CAS templates?

cmd/cstorvolumeclaim/cstorvolumeclaim.go
```diff
@@ -437,3 +442,27 @@ func randomizePoolList(list *apis.CStorPoolInstanceList) *apis.CStorPoolInstance
 
 	return res
 }
+
+// CVCWithVersionDetails build cvc resource with required version
+// details and volume references
+func CVCWithVersionDetails(
```
vishnuitta: this function is doing more than what the name of function suggest

cmd/cstorvolumeclaim/cstorvolumeclaim.go
prateekpandey14: Yes

cmd/cstorvolumeclaim/cstorvolumeclaim.go
prateekpandey14: Here `status.capacity` is getting populated as part of initial volume creation once the everthing got successfully created, which is helpful in case of resize compare conditions.
With this we will have a full resource instead of having it at the time of resize .

cmd/cspc-operator/app/handler.go
prateekpandey14: I am thinking to take this func below as this func can return nil cspc resource in case of any error, we should be breaking the provisioning flow IMO ?

cmd/cstorvolumeclaim/cstorvolumeclaim.go
prateekpandey14: what about  `WithVersionAndReferenceDetails` ?

cmd/cspc-operator/app/handler.go
prateekpandey14: taken return object

pkg/cstorvolumeclaim/v1alpha1/build.go
mittachaitu: This also good only intentions we shouldn't maintain capacity as a string but we can have capacity as `resource.Quantity` or `corev1.ResourceList`.
maintaining: 
`resource.Quantity`  will be useful only to store storage capacity
`corev1.ResourceList` can have multiple fields as below (depends on usecase) 
```
	// CPU, in cores. (500m = .5 cores)
	ResourceCPU ResourceName = "cpu"
	// Memory, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
	ResourceMemory ResourceName = "memory"
	// Volume size, in bytes (e,g. 5Gi = 5GiB = 5 * 1024 * 1024 * 1024)
	ResourceStorage ResourceName = "storage"
	// Local ephemeral storage, in bytes. (500Gi = 500GiB = 500 * 1024 * 1024 * 1024)
	// The resource name for ResourceEphemeralStorage is alpha and it can change across releases.
	ResourceEphemeralStorage ResourceName = "ephemeral-storage"
```

cmd/cstorvolumeclaim/controller.go
```diff
@@ -263,11 +263,12 @@ func (c *CVCController) createVolumeOperation(cvc *apis.CStorVolumeClaim) (*apis
...
 	}
 
-	// update the cstorvolume reference, phase as "Bound" and desired
-	// capacity
-	cvc.Spec.CStorVolumeRef = volumeRef
-	cvc.Status.Phase = apis.CStorVolumeClaimPhaseBound
-	cvc.Status.Capacity = cvc.Spec.Capacity
+	// update the cstorvolume reference, phase as "Bound", desired
+	// capacity and version details
+	cvc, err = CVCWithVersionAndRefernceDetails(cvc, volumeRef)
```
vishnuitta: spelling mistake

cmd/cspc-operator/app/handler.go
```diff
@@ -355,3 +364,58 @@ func (pc *PoolConfig) removeSPCFinalizerOnAssociatedBDC() error {
 
 	return nil
 }
+
+// populateVersion assigns VersionDetails for old cspc object and newly created
+// cspc
+func (c *Controller) populateVersion(cspc *apis.CStorPoolCluster) (*apis.CStorPoolCluster, error) {
+	if cspc.VersionDetails.Current == "" {
```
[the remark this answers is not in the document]
vishnuitta: fine.. no issues

cmd/cstorvolumeclaim/cstorvolumeclaim.go
```diff
@@ -437,3 +442,27 @@ func randomizePoolList(list *apis.CStorPoolInstanceList) *apis.CStorPoolInstance
...
 }
+
+// CVCWithVersionAndRefernceDetails build cvc resource with required version
+// details, volume references and status capacity
+func CVCWithVersionAndRefernceDetails(
+	claim *apis.CStorVolumeClaim,
+	volumeRef *corev1.ObjectReference,
+) (*apis.CStorVolumeClaim, error) {
+
+	cvcObj, err := cvc.BuildFrom(claim).
```
vishnuitta: we seem to be updating the claim that came in as input.. is that expected behavior of this fn?
We don't need to build.. we can just do update if thats the case..
we probably don't need function.. you can just do
`cvc.BuildFrom(claim).WithClaimRef(volumeRef).WithStatusPhase(apis.CStorVolumeClaimPhaseBound).WithNewVersion(version.GetVersion()).WithDependentsUpgrade()`

cmd/cspc-operator/app/handler.go
```diff
@@ -119,6 +122,12 @@ func (c *Controller) syncCSPC(cspcGot *apis.CStorPoolCluster) error {
 		return nil
 	}
 
+	cspcGot, err := c.populateVersion(cspcGot)
+	if err != nil {
+		klog.Errorf("failed to add versionDetails to CSPC %s:%s", cspcGot.Name, err.Error())
+		return err
```
vishnuitta: you want to return `nil`

cmd/cstorvolumeclaim/cstorvolumeclaim.go
vishnuitta: Let CSI create the CVC CR with versionDetails, so that, we don't need changes here

cmd/cstorvolumeclaim/controller.go
prateekpandey14: removed this code block as we are setting the version details from CSI code side

cmd/cspc-operator/app/handler.go
prateekpandey14: yes, changed

cmd/cstorvolumeclaim/cstorvolumeclaim.go
prateekpandey14: remove this from maya .. will be added from CSI side

PR #1518, opened 2019-11-12
fix(upgrade): handle errors for upgradeTask CR

shubham14bajpai:

This PR fixes wrong error reporting in different cases: 
 - When upgradeTask CR is not present then it should only return the actual error from the job.
 - When the CR is present it checks for backoff limit via env `POD_NAME` which is mentioned when the upgradeTask CR is created before the job is launched. In this case both the actual error and if the backoff limit error should be reported.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/upgrade/upgrader/exec.go
+++ b/pkg/upgrade/upgrader/exec.go
@@ -20,6 +20,8 @@ import (
 	"bytes"
 	"strings"
 
+	"k8s.io/klog"
+
 	apis "github.com/openebs/maya/pkg/apis/openebs.io/upgrade/v1alpha1"
 	csp "github.com/openebs/maya/pkg/cstor/pool/v1alpha3"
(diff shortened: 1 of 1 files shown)


pkg/upgrade/upgrader/exec.go
```diff
@@ -122,9 +124,10 @@ func Exec(fromVersion, toVersion, kind, name,
 	}
 
 	if err != nil {
-		if utaskObj != nil {
+		if utaskObj != nil && isENVPresent {
 			backoffLimit, uerr := getBackoffLimit(openebsNamespace)
 			if uerr != nil {
+				klog.Error(err)
```
mittachaitu: I think `err` should be returned and uerr should be logged

shubham14bajpai: Makes sense. Done

PR #1544, opened 2019-12-10
fix(cstor-pool-mgmt): fix livenessprobe in cStor pool deployment

mittachaitu:

This PR fixes the liveness probe on cstor-pool container by adding `timeout` setting for command execution(run a command with a time limit). `timeout` will be helpful in a case when the disks are detached from the node and when liveness triggers command(zfs set... command) it will be hung forever and kubelet will not treat them as a failures. Kubelet also retries execute the same command after `timeoutSeconds` mentioned in the liveness probe. By triggering `timeout 30 zfs set io.openebs:livenesstimestamp="$(date +%s)" cstor-<pool_name>` will kill the process if it exceeds more than 30 seconds and returns non-zero exit status.

**Note:**
1. When the disks are detached from the node cStor pool container will restart after 480 seconds i.e 8 minutes(tested in 1.4.8 Kubernetes version).
 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #https://github.com/openebs/openebs/issues/2852

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/install/v1alpha1/cstor_pool.go
+++ b/pkg/install/v1alpha1/cstor_pool.go
@@ -260,6 +260,8 @@ spec:
                 command:
                 - /bin/sh
                 - -c
+                ## timeout 30 is added to exit the command forcefully with non-zero exit code if command takes
+                ## more than 30 seconds
                 - timeout 30 zfs set io.openebs:livenesstimestamp="$(date +%s)" cstor-$OPENEBS_IO_CSTOR_ID
               failureThreshold: 3
               initialDelaySeconds: 300
--- a/pkg/install/v1alpha1/cstor_pool.go
+++ b/pkg/install/v1alpha1/cstor_pool.go
@@ -260,7 +260,7 @@ spec:
                 command:
                 - /bin/sh
                 - -c
-                - zfs set io.openebs:livenesstimestamp="$(date +%s)" cstor-$OPENEBS_IO_CSTOR_ID
+                - timeout 30 zfs set io.openebs:livenesstimestamp="$(date +%s)" cstor-$OPENEBS_IO_CSTOR_ID
               failureThreshold: 3
               initialDelaySeconds: 300
               periodSeconds: 10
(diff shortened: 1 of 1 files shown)


vishnuitta:
@singhmeghna79 would you be able to review this change and also try out if possible?

pkg/install/v1alpha1/cstor_pool.go
```diff
@@ -260,7 +260,7 @@ spec:
                 command:
                 - /bin/sh
                 - -c
-                - zfs set io.openebs:livenesstimestamp="$(date +%s)" cstor-$OPENEBS_IO_CSTOR_ID
+                - timeout 30 zfs set io.openebs:livenesstimestamp="$(date +%s)" cstor-$OPENEBS_IO_CSTOR_ID
```
mynktl: Let's add a comment here about timeout value.

mittachaitu: added

sonasingh46:
A similar fix needs to be propagated to the cspi mgmt YAML spec that can be done in a separate PR.

pkg/install/v1alpha1/cstor_pool.go
```diff
@@ -260,7 +260,9 @@ spec:
                 command:
                 - /bin/sh
                 - -c
-                - zfs set io.openebs:livenesstimestamp="$(date +%s)" cstor-$OPENEBS_IO_CSTOR_ID
+                ## timeout 30 is added to exit the command forcefully with non-zero exit code if command takes
```
kmova: Make this 300s to avoid false positives.

mittachaitu: Updated settings.

mittachaitu:
> A similar fix needs to be propagated to the cspi mgmt YAML spec that can be done in a separate PR.

Handled in PR #1546

PR #1559, opened 2019-12-17
refact(cvc): move cvc operator to its own deployment

prateekpandey14:

Changes remove the cstorvolumeclaim controller go-routines from the maya-apiserver and will be deployed as its own pod in k8s cluster.

This changes required to be done as cvc controller is specific to CStor CSI volume provisioning and there is no depdencies in Maya-apiserver.
As part of moving further with buiding new controllers/operators specific to each components and there specific repos it will easier to maintain.

This requires a new deployment file to be added as part of cstor-operator, PR https://github.com/openebs/openebs/pull/2868


GNUmakefile
```diff
@@ -422,6 +425,17 @@ cspc-operator-debug-image:
...
 	@rm buildscripts/${CSPC_OPERATOR_DEBUG}/${CSPC_OPERATOR}
 
+.PHONY: cvc-operator-image
+cvc-operator-image:
+	@echo "----------------------------"
+	@echo -n "--> cvc-operator image "
+	@echo "${HUB_USER}/${CVC_OPERATOR_REPO_NAME}:${IMAGE_TAG}"
+	@echo "----------------------------"
+	@PNAME=${CVC_OPERATOR} CTLNAME=${CVC_OPERATOR} sh -c "'$(PWD)/buildscripts/build.sh'"
+	@cp bin/${CVC_OPERATOR}/${CVC_OPERATOR} buildscripts/cvc-operator/
```
mittachaitu: can we remove cvc-operator? and use ${CVC_OPERATOR}

prateekpandey14: yes we can use , but i just follow the other tags

GNUmakefile
```diff
@@ -329,6 +333,17 @@ cspc-operator-debug-image:
 	@cd buildscripts/${CSPC_OPERATOR_DEBUG} && sudo docker build -t ${HUB_USER}/${CSPC_OPERATOR_REPO_NAME}:inject --build-arg BUILD_DATE=${BUILD_DATE} .
 	@rm buildscripts/${CSPC_OPERATOR_DEBUG}/${CSPC_OPERATOR}
 
+.PHONY: cvc-operator-image
```
kmova: Move this into its own .mk file

prateekpandey14: Done

PR #1583, opened 2020-01-09
fix(cStor, deployments): add OpenEBS base directory in deployments

mittachaitu:

This PR mounts the `/var/openebs` of the container into  OPENEBS_IO_BASE_DIR/<cstor_target/pool>/<resource_name> on the host path.
These changes are made to persist core files on the host path.

**How Users Can Configure**
Users can specify the value of OPENEBS_IO_BASE_DIR in Maya deployment.
While creating cStor pool/volume corresponding component will read the ENV value
and use base_dir for mounting. If ENV doesn't exist base_dir will mount `/var/openebs`.

Below are PR's to support dumping core in /var/openebs/core inside a container:
cStor: https://github.com/openebs/cstor/pull/284
istgt: https://github.com/openebs/istgt/pull/298

These changes are for SPC type of provisioning subsequent PRs will be raised for CSPC.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/install/v1alpha1/cstor_pool.go
+++ b/pkg/install/v1alpha1/cstor_pool.go
@@ -44,6 +44,13 @@ spec:
   # nothing exists at the given path i.e. an empty directory will be created.
   - name: HostPathType
     value: DirectoryOrCreate
+  # PersistentStoragePath is base directory to store lock, sock and cache and core files.
+  # PersistentStoragePath should be used only to mount inside a container.
+  - name: PersistentStoragePath
+    value: "/var/openebs"
+  # OpenebsBaseDir is a hostPath directory to store process files on host machine
+  - name: OpenebsBaseDir
+    value: {{env "OPENEBS_IO_BASE_DIR" | default "/var/openebs"}}
   # SparseDir is a hostPath directory where to look for sparse files
   - name: SparseDir
     value: {{env "OPENEBS_IO_CSTOR_POOL_SPARSE_DIR" | default "/var/openebs/sparse"}}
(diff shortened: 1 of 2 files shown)


pkg/install/v1alpha1/cstor_pool.go
```diff
@@ -44,6 +44,13 @@ spec:
   # nothing exists at the given path i.e. an empty directory will be created.
   - name: HostPathType
     value: DirectoryOrCreate
+  # PersistentStoragePath is base directory to store lock, sock and cache and core files.
```
vishnuitta: is this comment correct?

mittachaitu: I will update it I picked it from different PR.

pkg/install/v1alpha1/cstor_volume.go
```diff
@@ -579,6 +586,8 @@ spec:
               mountPath: /var/run
             - name: conf
               mountPath: /usr/local/etc/istgt
+            - name: storagepath
+              mountPath: {{ .Config.PersistentStoragePath.value }}
```
pawanpraka1: why PersistentStoragePath *variable* is needed? it  has to be "/var/openebs" all the time, other wise istgt image needs to be regenerated with the new mount path? isn't it?

mittachaitu: Yes, If that variable changed then istgt image also needs to be regenerated. So one shouldn't touch that `PersistentStoragePath` variable(Made comments above that variable).
We are using in different places so made variable to make it use.

mittachaitu: I will hard code it

PR #1590, opened 2020-01-22
chore(log): remove v2 verbosity for csp selection in case of overprovisioning

sonasingh46:
The PR removes v2 verbosity for CSP selection in case of overprovisioning and put it to regular info.
Signed-off-by: Ashutosh Kumar <ashutosh.kumar@mayadata.io>


vishnuitta:
Event or message need to print volume name as well and should not log every failure in selecting CSP.
Hence, closing this PR.

PR #1591, opened 2020-01-22
feat(cvc,csi): pre provisioning and prioritize pool for replica scheduling

prateekpandey14:

changes includes:
- adds support of pre provisioning cstor volumes instead of always
  waitForFirstConsumer
- adds support of prioritize cstor pool based required for StatefulSet
  replica affinity scenarios
- Updated the volume policy to have new parameter to enable-disable
  replicaAffinity


pkg/apis/openebs.io/v1alpha1/cstorvolume_policy.go
```diff
@@ -38,6 +38,9 @@ type CStorVolumePolicy struct {
 
 // CStorVolumePolicySpec ...
 type CStorVolumePolicySpec struct {
+	// replicaAffinity is set to true then volume resources need to be
+	// distributed across the pool instances
+	ReplicaAffinity bool `json:"replicaAffinity"`
```
payes: Let's update the correct explanation for this variable

cmd/cvc-operator/controller/cvc_controller.go
```diff
@@ -304,6 +305,17 @@ func (c *CVCController) getVolumePolicy(
 	return volumePolicy, nil
 }
 
+// isReplicaAffinityEnabled checks if replicaAffinity has been enabled using
+// cstor volume policy
+func (c *CVCController) isReplicaAffinityEnabled(cvc *apis.CStorVolumeClaim) bool {
+	policyName := cvc.Annotations[string(apis.VolumePolicyKey)]
+	volumePolicy, err := c.getVolumePolicy(policyName, cvc)
```
payes: Let's return default policy if policy is not provided by user in storage class

cmd/cvc-operator/controller/cstorvolumeclaim.go
```diff
@@ -350,8 +350,15 @@ func distributeCVRs(
...
 		usablePoolList = getUsablePoolList(volume.Name, poolList)
 	}
+
 	// randomizePoolList to get the pool list in random order
 	usablePoolList = randomizePoolList(usablePoolList)
+
+	// prioritized pool instances matched to the given
+	// nodeName in case of replica affinity is enabled via cstor volume policy
+	if c.isReplicaAffinityEnabled(claim) {
+		usablePoolList = prioritizedPoolList(claim.Publish.NodeID, usablePoolList)
```
mittachaitu: This will cause an imbalance in the pool i.e I have a storage class without topology and created an application using that storage class with replica antiaffinity now I think CSI-driver will always pick the first node from that list this will cause imbalance the pool.
cc: @payes

cmd/cvc-operator/controller/cstorvolumeclaim.go
```diff
@@ -512,6 +519,19 @@ func getUsablePoolListForClone(pvName, srcPVName string, poolList *apis.CStorPoo
 	return usablePoolList
 }
 
+// prioritizedPoolList prioritized pool instance name matched to the given
+// nodeName in case of replica affinity is enabled via volume policy
+func prioritizedPoolList(nodeName string, list *apis.CStorPoolInstanceList) *apis.CStorPoolInstanceList {
+	for i, pool := range list.Items {
+		if pool.Spec.HostName != nodeName {
```
mittachaitu: Here, we are checking nodeName with HostName. Will it work if nodeName differs from hostName?

We are populating nodeName
https://github.com/openebs/maya/blob/5d066c6fd24f0dbe9a463096929b83f741731a5b/pkg/cstor/poolinstance/v1alpha3/build.go#L192

cmd/cvc-operator/controller/cstorvolumeclaim.go
```diff
@@ -512,6 +519,19 @@ func getUsablePoolListForClone(pvName, srcPVName string, poolList *apis.CStorPoo
...
+			continue
+		}
+		list.Items[0], list.Items[i] = list.Items[i], list.Items[0]
```
mittachaitu: nit: if it is already at 0th indexes we no need to swap.

cmd/cvc-operator/controller/cstorvolumeclaim.go
```diff
@@ -512,6 +519,19 @@ func getUsablePoolListForClone(pvName, srcPVName string, poolList *apis.CStorPoo
...
+
```
mittachaitu: add break

cmd/cvc-operator/controller/cstorvolumeclaim.go
mittachaitu: We have to fix in CSPI spec

cmd/cvc-operator/controller/cvc_controller.go
prateekpandey14: Default policy is not part of this PR changes

pkg/apis/openebs.io/v1alpha1/cstorvolume_policy.go
mittachaitu: Can it be under `Replica`?

cmd/cvc-operator/controller/cstorvolumeclaim.go
```diff
@@ -350,8 +350,15 @@ func distributeCVRs(
 	} else {
 		usablePoolList = getUsablePoolList(volume.Name, poolList)
```
vishnuitta: this 'if else' doesn't look good coding style

cmd/cvc-operator/controller/cstorvolumeclaim.go
vishnuitta: this comment need to be fixed

vishnuitta: sorry.. not this.. but, this needs to get into documentation

cmd/cvc-operator/controller/cstorvolumeclaim.go
vishnuitta: this need to be fixed

cmd/cvc-operator/controller/cvc_controller.go
```diff
@@ -271,6 +271,7 @@ func (c *CVCController) createVolumeOperation(cvc *apis.CStorVolumeClaim) (*apis
 	// update the cstorvolume reference, phase as "Bound" and desired
 	// capacity
 	cvc.Spec.CStorVolumeRef = volumeRef
+	cvc.Spec.Policy = volumePolicy.Spec
```
vishnuitta: this need to be moved up after getting policy..

cmd/cvc-operator/controller/cvc_controller.go
vishnuitta: why are doing getVolumePolicy multiple times?

pkg/apis/openebs.io/v1alpha1/cstor_volume_claim.go
```diff
@@ -83,6 +85,11 @@ const (
 	//CStorVolumeClaimPhaseFailed indiacates that the cstorvolume provisioning
 	//has failed
 	CStorVolumeClaimPhaseFailed CStorVolumeClaimPhase = "Failed"
+
+	//CStorVolumeClaimPhaseReconcile indicates that the changes currently
+	//reconclied based on the cstor volume policy changes to achieve the desired
+	//state
+	CStorVolumeClaimPhaseReconcile CStorVolumeClaimPhase = "Reconile"
```
vishnuitta: it seems - we are not using this

pkg/apis/openebs.io/v1alpha1/cstorvolume_policy.go
vishnuitta: This doesn't go under Replica.. probably, new section like `VolumeConfig` or `Provisioner`?

pkg/apis/openebs.io/v1alpha1/cstor_volume_claim.go
prateekpandey14: We will be using the for the reconcile purpose later, should i remove it ?

cmd/cvc-operator/controller/cvc_controller.go
prateekpandey14: We can pass the policy as an arg to caller , will change that

prateekpandey14: Some changes are not part of this PR, as they are coming with reconcile changes and default policy changes

cmd/cvc-operator/controller/cvc_controller.go
prateekpandey14: Some changes are not part of this PR, as they are coming with reconcile changes and default policy changes

pkg/apis/openebs.io/v1alpha1/cstorvolume_policy.go
mittachaitu: FIne with it

cmd/cvc-operator/controller/cvc_controller.go
prateekpandey14: Done, removed the redundant call in this PR itself.

cmd/cvc-operator/controller/cvc_controller.go
```diff
@@ -258,7 +258,7 @@ func (c *CVCController) createVolumeOperation(cvc *apis.CStorVolumeClaim) (*apis
 	}
 
 	klog.V(2).Infof("creating cstorvolume replica resource")
-	err = c.distributePendingCVRs(cvc, cvObj, svcObj)
+	err = c.distributePendingCVRs(cvc, cvObj, svcObj, volumePolicy)
```
vishnuitta: isn't the check for `cvc.Publish.NodeID` being NULL need to be removed in caller of createVolumeOperation?

cmd/cvc-operator/controller/cvc_controller.go
vishnuitta: (not just related to this PR) we are modifying the objects from store.. is this fine?

pkg/apis/openebs.io/v1alpha1/cstor_volume_claim.go
vishnuitta: lets remove and add when needed

pkg/apis/openebs.io/v1alpha1/cstorvolume_policy.go
vishnuitta: we are not moving it to new section?

cmd/cvc-operator/controller/cstorvolumeclaim.go
vishnuitta: how this got fixed?

cmd/cvc-operator/controller/cvc_controller.go
prateekpandey14: `NodeID` will not be null, we will be having the picked one of the preferred node name everytime from the driver side as, it will available during volume create request at csi driver side.

cmd/cvc-operator/controller/cvc_controller.go
prateekpandey14: Yes, that's k8s does that for there resources to have some default values

pkg/apis/openebs.io/v1alpha1/cstor_volume_claim.go
prateekpandey14: done

cmd/cvc-operator/controller/cvc_controller.go
vishnuitta: how can CSI driver fill the NodeID when there is no `allowedTopologies` or `WaitForFirstConsumer` is set in storageclass?

PR #1599, opened 2020-01-30
fix(new cStor deployments): add OpenEBS base directory in new CStor deployments

mittachaitu:

This PR is a continuation of PR: https://github.com/openebs/maya/pull/1583 and new deployments strictly follow to store component related files in one path of the host machine and that path called OPENEBS_IO_BASE_DIR.
**How Users Can Configure:**
If users need to configure base directory he has to set on OPENEBS_IO_BASE_DIR Env in cspc-operator & CVC-operator.

What files stored in the host machine:
- When CSPC is deployed pool pod will generate lock file, sock file, core files, and pool cache file inside the container and the container path is mounted on  **BASE_DIR/cstor-pool/<cspc_name>** host path.
- When cStor volume is deployed target pod will store core files inside the container and the container path is mounted on **BASE_DIR/cstor-target/<volume_name>**  host path.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/cmd/cspc-operator/app/storagepool_create.go
+++ b/cmd/cspc-operator/app/storagepool_create.go
@@ -21,6 +21,7 @@ import (
 
 	apis "github.com/openebs/maya/pkg/apis/openebs.io/v1alpha1"
 	apiscsp "github.com/openebs/maya/pkg/cstor/poolinstance/v1alpha3"
+	env "github.com/openebs/maya/pkg/env/v1alpha1"
 	container "github.com/openebs/maya/pkg/kubernetes/container/v1alpha1"
 	deploy "github.com/openebs/maya/pkg/kubernetes/deployment/appsv1/v1alpha1"
 	pts "github.com/openebs/maya/pkg/kubernetes/podtemplatespec/v1alpha1"
@@ -62,6 +63,10 @@ var (
 		corev1.VolumeMount{
 			Name:      "udev",
 			MountPath: "/run/udev",
+		},
+		corev1.VolumeMount{
+			Name:      "storagepath",
+			MountPath: "/var/openebs/cstor-pool",
 		},
 	}
 	// hostpathType represents the hostpath type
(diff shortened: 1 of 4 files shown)


cmd/cspc-operator/app/storagepool_create.go
```diff
@@ -174,7 +179,7 @@ func (pc *PoolConfig) GetPoolDeploySpec(cspi *apis.CStorPoolInstance) (*appsv1.D
 					volume.NewBuilder().
 						WithName("tmp").
 						WithHostPathAndType(
-							getSparseDirPath()+"/shared-"+pc.AlgorithmConfig.CSPC.Name,
+							env.GetOpenebsBaseDirPath()+"/cstor-pool/"+pc.AlgorithmConfig.CSPC.Name,
```
vishnuitta: migration from SPC to CSPC need to be taken care due to this cc: @shubham14bajpai

shubham14bajpai: Sure @vishnuitta. Just one small question, are we going to ask users to come to the latest version of SPC before migration? If yes then upgrade will take care of this automatically.

vishnuitta: yes.. we should do migration from latest openebs version

vishnuitta:
> should we remove '-target' from this line of `path: {{ .Config.TargetDir.value }}/shared-{{ .Volume.owner }}-target` from CAS templates?

not from above line.. but, the one in storagepath

mittachaitu:
> should we remove '-target' from this line of `path: {{ .Config.TargetDir.value }}/shared-{{ .Volume.owner }}-target` from CAS templates?

No, @vishnuitta if we remove then we have to perform cleanup of the old directory on host path which will be redundant work(since we are planned to do it during migration).

PR #1600, opened 2020-01-31
fix(jiva-cleanup): allow cleanup jobs to run as previleged mode

utkarshmani1997:

This commit fix the issue where scrub jobs were failing on openshift platform
or where non previleged pods are not allowed to access/modify hostpath.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes # https://github.com/openebs/openebs/issues/2893

none
**Checklist:**
- [x] Fixes openebs/openebs#2893
- [ ] Commit has unit tests

### DIFF
--- a/pkg/install/v1alpha1/jiva_volume.go
+++ b/pkg/install/v1alpha1/jiva_volume.go
@@ -1390,6 +1390,8 @@ spec:
               type: ""
           containers:
           - name: sjr
+            securityContext:
+              privileged: true
             image: {{ .Config.ScrubImage.value }}
             command:
             - sh


pkg/install/v1alpha1/jiva_volume.go
```diff
@@ -1390,6 +1390,8 @@ spec:
               type: ""
           containers:
           - name: sjr
+            securityContext:
```
vishnuitta: why it needs privileged access?

kmova: @utkarshmani1997 can you launch it with serviceAccount like replica pods.

utkarshmani1997: @vishnuitta @kmova there is no service account for the replica pods. It runs as privileged pod. Also this doesn't require service account, service account is required only when any pod/deployment/sts/ds in a cluster require access to do some operation of the kubernetes resource. 
In this case we need to have PSP which will provide access to the configured resources. So this needs to be introduced in replica as well.

utkarshmani1997: Ref: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#policy-reference
Please have a look at allowHostPath section

utkarshmani1997: or scc needs to be created to allowHostPath in openshift clusters
Ref: https://docs.openshift.com/enterprise/3.0/admin_guide/manage_scc.html#use-the-hostpath-volume-plugin
cc: @nsathyaseelan

kmova: understood. thanks @utkarshmani1997

PR #1611, opened 2020-02-11
fix(cspc): replace the spc label with cspc label

shubham14bajpai:

This PR replaces the spc label with cspc label in the cspc-operator. If the finalizers are removed from the cspi and for some unknown reason the deletion is stuck then in that case the cspc should not be deleted before cspi deletion.

### DIFF
--- a/cmd/cspc-operator/app/handler.go
+++ b/cmd/cspc-operator/app/handler.go
@@ -332,7 +332,7 @@ func (pc *PoolConfig) removeCSPCFinalizer() error {
 	}
 
 	cspList, err := apiscsp.NewKubeClient().List(metav1.ListOptions{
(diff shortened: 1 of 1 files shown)


cmd/cspc-operator/app/handler.go
```diff
@@ -332,7 +332,7 @@ func (pc *PoolConfig) removeCSPCFinalizer() error {
 	}
 
 	cspList, err := apiscsp.NewKubeClient().List(metav1.ListOptions{
-		LabelSelector: string(apis.StoragePoolClaimCPK) + "=" + pc.AlgorithmConfig.CSPC.Name,
+		LabelSelector: string(apis.CStorPoolClusterCPK) + "=" + pc.AlgorithmConfig.CSPC.Name,
```
mittachaitu: :+1:  Good one

PR #1615, opened 2020-02-13
fix(exporter): Handle pool sync time metrics collection gracefully

singhmeghna79:
Signed-off-by: Meghna Singh <singhmegna79@gmail.com>

This PR fixes a bug in pool sync time metrics collector in maya exporter. If last request is in progress, don't queue the current request during pool last Sync time metrics collection.

### DIFF
--- a/cmd/maya-exporter/app/collector/zvol/pool_synctime.go
+++ b/cmd/maya-exporter/app/collector/zvol/pool_synctime.go
@@ -51,6 +51,10 @@ func (p *poolMetrics) setRequestToFalse() {
 	p.Unlock()
 }
 
+func (p *poolMetrics) isRequestInProgress() bool {
+	return p.request
+}
+
 // collectors returns the list of the collectors
 func (p *poolMetrics) collectors() []prometheus.Collector {
 	return []prometheus.Collector{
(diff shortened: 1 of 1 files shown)


singhmeghna79:
> Is this the way we are doing in other exporters as well @singhmeghna79 ?
Yes @avishnu

cmd/maya-exporter/app/collector/zvol/pool_synctime.go
```diff
@@ -110,6 +114,10 @@ func (p *poolMetrics) get() *poolfields {
 // Collect is implementation of prometheus's prometheus.Collector interface
 func (p *poolMetrics) Collect(ch chan<- prometheus.Metric) {
 	p.Lock()
+	if p.isRequestInProgress() {
```
avishnu: Are we trying to prevent a deadlock from happening with this code?
Out of curiosity..
How does it get a lock when another request is in progress?

utkarshmani1997: It's only protecting request flag.

akhilerm: @avishnu . Its not necessarily a deadlock. we are preventing requests from being queuing up.
We take the lock on the request flag, check if a request is in progress or not and release the lock. This will help us take lock only for a short time, and only for the single request flag, instead of the complete metric struct.

Issue #33, opened 2017-03-14
support for install/setup of etcd master node


kmova:
maya cli should include an option to setup etcd-master.  The command could be:
```
maya setup-etcd-master [args....]
```
arguments could include the listen/expose/advertise ip addresses, peer etcd-master addresses, certificates etc.,  Use the following steps as a sample for installing the etcd. 

```
sudo apt-get install etcd
```

Config parameters could be:
```
ETCD_NAME=k8s-master-1
ETCD_DATA_DIR="/var/lib/etcd"
ETCD_LISTEN_PEER_URLS="https://172.28.128.10:2380"
ETCD_LISTEN_CLIENT_URLS="https://172.28.128.10:2379,http://localhost:2379"
ETCD_INITIAL_ADVERTISE_PEER_URLS="https://172.28.128.10:2380"
ETCD_INITIAL_CLUSTER="k8s-master-1=https://172.28.128.10:2380"
ETCD_INITIAL_CLUSTER_STATE="new"
ETCD_INITIAL_CLUSTER_TOKEN="etcd-cluster-0"
ETCD_ADVERTISE_CLIENT_URLS="https://172.28.128.10:2379"
ETCD_CERT_FILE="/etc/etcd/client.crt"
ETCD_KEY_FILE="/etc/etcd/client.key"
ETCD_TRUSTED_CA_FILE="/etc/etcd/ca.crt"
ETCD_PEER_CERT_FILE="/etc/etcd/client.crt"
ETCD_PEER_KEY_FILE="/etc/etcd/client.key"
ETCD_PEER_TRUSTED_CA_FILE="/etc/etcd/ca.crt"
```

PR #1642, opened 2020-03-31
feat(apis): add new fields into BlockDevice

akhilerm:
Signed-off-by: Akhil Mohan <akhil.mohan@mayadata.io>

- add DriveType and BlockSize fields into BlockDevice spec

- These are the same changes as in PR [#33](https://github.com/openebs/api/pull/33)


kmova:
Travis has passed, but the notification has not been updated.

PR #1663, opened 2020-04-10
fix(upgarde): scale down jiva contoller deployment only if not upgraded

shubham14bajpai:

This PR fixes the issue which might occur if the upgrade job is run after successful completion. It might scale down the controller which will not be patched and remained scaled-down forever.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/upgrade/upgrader/jiva_upgrade.go
+++ b/pkg/upgrade/upgrader/jiva_upgrade.go
@@ -517,20 +517,32 @@ func (j *jivaVolumeOptions) replicaUpgrade(pvName, openebsNamespace string) erro
 	}
 	statusObj.Phase = utask.StepErrored
 
-	// Scale down controller deployment only if the controller is not upgraded.
-	if j.controllerObj.version == currentVersion {
-		// Scaling down controller ensures no I/O occurs
-		// which make volume to come in RW mode early
-		err = scaleDeploy(j.controllerObj.name, j.ns, ctrlDeployLabel, 0)
-		if err != nil {
-			statusObj.Message = "failed to get scale down target deployment"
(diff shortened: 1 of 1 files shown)


pkg/upgrade/upgrader/jiva_upgrade.go
```diff
@@ -517,17 +517,20 @@ func (j *jivaVolumeOptions) replicaUpgrade(pvName, openebsNamespace string) erro
...
-	// Scaling down controller ensures no I/O occurs
-	// which make volume to come in RW mode early
-	err = scaleDeploy(j.controllerObj.name, j.ns, ctrlDeployLabel, 0)
-	if err != nil {
-		statusObj.Message = "failed to get scale down target deployment"
-		statusObj.Reason = strings.Replace(err.Error(), ":", "", -1)
-		j.utaskObj, uerr = updateUpgradeDetailedStatus(j.utaskObj, statusObj, openebsNamespace)
-		if uerr != nil && isENVPresent {
-			return uerr
+	// Scale down controller deployment only if the controller is not upgraded.
```
vishnuitta, line 520: should we consider returning from this function if controller's version is upgradeVersion?

shubham14bajpai: In that case, success status will not be patched for REPLICA_UPGRADE step in the upgradeTask CR and will be stuck in pending state.

vishnuitta: ok.. we can patch the upgradeTask CR also at the end

shubham14bajpai: done

PR #1678, opened 2020-05-03
fix(maya-exporter): fix duplicate metrics for openebs_zpool_last_sync_time

slalwani97:
This PR fixes the generation of duplicate metrics of openebs_zpool_last_sync_time metric.

**What this PR does?**:
This PR makes the labeled `pool` value as a pod name, which means the label will be the same in case of error and non-error. So that it does not lead to two different metrics. It also does not set the last_sync_time value in case of error so as to retain the previous metric value.

### DIFF
--- a/changelogs/unreleased/1678-slalwani97
+++ b/changelogs/unreleased/1678-slalwani97
@@ -1,0 +1,1 @@
+fix(maya-exporter): fix duplicate metrics generation after pool pod restart or any error.
--- a/changelogs/unreleased/1678-slalwani97
+++ b/changelogs/unreleased/1678-slalwani97
@@ -1,1 +1,1 @@
-fix(maya-exporter): fix duplicate metrics generation after pool pod restart or any error.
+fix(maya-exporter): duplicate metrics generation after pool pod restart or any error.
--- a/cmd/maya-exporter/app/collector/zvol/pool_synctime_test.go
+++ b/cmd/maya-exporter/app/collector/zvol/pool_synctime_test.go
@@ -16,6 +16,7 @@ package zvol
 
 import (
 	"fmt"
+	"os"
 	"testing"
 
 	types "github.com/openebs/maya/pkg/exec"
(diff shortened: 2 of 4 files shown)


cmd/maya-exporter/app/collector/zvol/pool_synctime_test.go
```diff
@@ -37,9 +37,9 @@ func TestPoolSyncTimeMetricCollector(t *testing.T) {
 			zpoolLastSyncTimeOutput: `cstor-c6f17743-e5d7-11e9-b673-42010a800112      io.openebs:livenesstimestamp    1570625404      local
 			cstor-c6f17743-e5d7-11e9-b673-42010a800112/pvc-053647ca-e5d8-11e9-b673-42010a800112     io.openebs:livenesstimestamp    1570625404      inherited from cstor-c6f17743-e5d7-11e9-b673-42010a800112`,
 			expectedOutput: []string{
-				`openebs_zpool_last_sync_time{pool="cstor-c6f17743-e5d7-11e9-b673-42010a800112"} 1.570625404e\+09`,
-				`openebs_zpool_state_unknown{pool="cstor-c6f17743-e5d7-11e9-b673-42010a800112"} 0`,
-				`openebs_zpool_sync_time_command_error{pool="cstor-c6f17743-e5d7-11e9-b673-42010a800112"} 0`,
+				`openebs_zpool_last_sync_time{pool=""} 1.570625404e\+09`,
```
kmova, line 40: can you modify this test to use the hostname?

Prior to running the test you can modify the os.evn hostname to what will typically be seen within the pool pod and then set it back to whatever was the original value after the test is done. 

Example: https://github.com/openebs/maya/blob/master/cmd/mayactl/app/command/volume_stats_test.go#L118

kmova:
Also @slalwani97 -- can you follow the new guidelines on how to include the changelog into the PR itself. https://github.com/openebs/maya/blob/master/code-standard.md#adding-a-changelog

changelogs/unreleased/1678-slalwani97
```diff
@@ -0,0 +1 @@
+fix(maya-exporter): fix duplicate metrics generation after pool pod restart or any error.
```
kmova, line 1: n_: no need to mention "fix" again in the changelog as it is already present in type.

cmd/maya-exporter/app/collector/zvol/pool_synctime_test.go
```diff
@@ -104,6 +112,8 @@ func TestPoolSyncTimeMetricCollector(t *testing.T) {
 			regex := mockServer.BuildRegex(out)
 			vol := NewPoolSyncMetric(tt.runner)
 			stop := make(chan struct{})
+			os.Setenv("HOSTNAME", tt.hostname)
+			defer os.Unsetenv("HOSTNAME")
```
kmova, line 116: probably better to set it back to whatever was the value prior to running this test.

changelogs/unreleased/1678-slalwani97
slalwani97: Sure, done.

cmd/maya-exporter/app/collector/zvol/pool_synctime_test.go
slalwani97: ok.

kmova:
@slalwani97 - can you run the `make test` on the code. Travis is failing due to go fmt changes.

PR #1693, opened 2020-05-12
feat(install): enable/disable crd installation

kmova:

When helm install for openebs was introduced,
there was no way for helm to install CRDs. So
api-server included the code to install the CRDs
as part of initialization.

With Helm 3, the CRDs can be installed via
helm. There have been some requests for moving
the CRD installations outside of the code for
cases where Kubernetes installers like Lokomotive
and others, would like granular control on
how the required objects are installed.

**What this PR does?**:

This commit adds an ENV variable that allows
users to enable/disable the CRD install code.

NA

_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [ ] Fixes #<issue number>
- [x] Commit has unit tests

### DIFF
--- a/pkg/install/v1alpha1/env.go
+++ b/pkg/install/v1alpha1/env.go
@@ -35,4 +35,12 @@ const (
 	//
 	// Default is "true"
 	CreateDefaultStorageConfig menv.ENVKey = "OPENEBS_IO_CREATE_DEFAULT_STORAGE_CONFIG"
+
+	// InstallCRD is the environment
+	// variable that flags if maya apiserver should install the CRDs
+	// As the installation moves towards helm 3, the responsibility of installing
+	// CRDs can be pushed to helm.
+	//
+	// Default is "true"
+	InstallCRD menv.ENVKey = "OPENEBS_IO_INSTALL_CRD"
 )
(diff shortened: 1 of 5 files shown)


akhilerm:
@kmova  Is the same set of changes required in NDM, for the installation of `BlockDevice` and `BlockDeviceClaim` CRDs?

kmova:
@akhilerm - probably makes sense to add for NDM as well.

pkg/install/v1alpha1/env.go
```diff
@@ -35,4 +35,12 @@ const (
...
 	// Default is "true"
 	CreateDefaultStorageConfig menv.ENVKey = "OPENEBS_IO_CREATE_DEFAULT_STORAGE_CONFIG"
+
+	// InstallCRD is the environment
+	// variable that flags if maya apiserver should install the CRDs
+	// As the installation moves towards helm 3, the responsibility of installing
+	// CRDs can be pushed to helm.
+	//
+	// Default is "true"
+	InstallCRD menv.ENVKey = "OPENEBS_IO_INSTALL_CRD"
```
prateekpandey14, line 45: can we name it like `InstallV1Alpha1CRDs` ?

kmova: Not sure I understand the reason? Will there be a case that we will allow installing other types of CRDs?

prateekpandey14: Ignore the above comment, i just got confused with ENV name going to expose to user

PR #1696, opened 2020-05-14
fix(webhook): cleanup old resources from 1.0.0 release

shubham14bajpai:

This PR handles the successful cleanup of old service, secret and webhookconfig for admission webhook in case of operator and helm due to non-uniform labels across different installations in v1.0.0 once they upgrade v1.10.0 version.

Above issue has been seen, when we install openebs v1.0.0 version using openebs-operator yaml 
and  upgrading the v1.0.0 version to the any greater version ( v1.1.0 onwards) using openebs-operator yaml (not with helm as helm maintain there set of labels in each resources, which take cares of upgrading/patching the resources).

Since we have changed the Webhook `FailurePolicy: Fail` , this causes failure for any configuration mistakes in webhook in older version v1.0.0.

Other version since v1.1.0 onwards doesn't have similar upgrade issue due to matching similar labels in subsequent versions and upgrade is taken care by the webhook itself internally as a preupgrade step.

 No

- Tested Upgrade using Helm (although helm doesn't have stale resource issue)
   - from 1.0.0-->1.5.0-->1.10.0 ( via using helm upgrade)
   - Provisioned cstor and jiva volumes
- Tested Upgrade using Operator yaml ( via just applying openebs-operator.yaml)
  - from 1.0.0-->1.5.0-->1.10.0
  - Provisioned cstor and jiva volumes

### DIFF
--- a/pkg/webhook/configuration.go
+++ b/pkg/webhook/configuration.go
@@ -491,7 +491,7 @@ func getOldService(openebsNamespace string) (*corev1.ServiceList, error) {
 	// fetch service 1.1.0 onwards based on label
 	svcList, err := svc.NewKubeClient(svc.WithNamespace(openebsNamespace)).List(metav1.ListOptions{LabelSelector: webhooksvcLabel})
 	if err != nil {
-		return nil, fmt.Errorf("failed to list old service: %s", err.Error())
+		return nil, err
 	}
 	// for 1.0.0 fetch service based on old name
 	oldSVC, err := svc.NewKubeClient(svc.WithNamespace(openebsNamespace)).Get(v100SVCName, metav1.GetOptions{})
@@ -512,7 +512,7 @@ func getOldSecret(openebsNamespace string) (*corev1.SecretList, error) {
 	secretList, err := secret.NewKubeClient(secret.WithNamespace(openebsNamespace)).
 		List(metav1.ListOptions{LabelSelector: webhookLabel})
 	if err != nil {
-		return nil, fmt.Errorf("failed to list old secret: %s", err.Error())
+		return nil, err
 	}
 	// for 1.0.0 fetch secret based on old name
 	oldSecret, err := secret.NewKubeClient(secret.WithNamespace(openebsNamespace)).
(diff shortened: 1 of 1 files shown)


pkg/webhook/configuration.go
```diff
@@ -486,10 +486,79 @@ func addCVCWithUpdateRule(config *v1beta1.ValidatingWebhookConfiguration) {
...
+	}
+	if err == nil {
+		secretList.Items = append(secretList.Items, *oldSecret)
+	}
+	return secretList, nil
+
+}
+
+func getOldConfig() (*v1beta1.ValidatingWebhookConfigurationList, error) {
+	v100HelmConfig := "openebs-validation-webhook-cfg"
```
kmova, line 531: This may not apply, as openebs via help can be installed with any other name, and then this would be "funnyname-validation-webhook-cfg"? 

Is there any other way to get this?

prateekpandey14: @kmova we have constant names in case of helm

shubham14bajpai: This is also fixed helm.
https://github.com/helm/charts/blob/b2bc5480b16bd29587cbb0a295b35b2306c81dc0/stable/openebs/templates/validationwebhook.yaml#L10

kmova: yes. if you don't give "--name  openebs" in install instruction, it installs with custom name. Please try it.

prateekpandey14: helm v1.0.0: https://github.com/helm/charts/blob/b2bc5480b16bd29587cbb0a295b35b2306c81dc0/stable/openebs/templates/validationwebhook.yaml

operator: https://github.com/openebs/charts/blob/master/docs/openebs-operator-1.0.0.yaml

kmova: you are right. I see the hardcoding. https://github.com/helm/charts/blob/b2bc5480b16bd29587cbb0a295b35b2306c81dc0/stable/openebs/templates/validationwebhook.yaml#L10

kmova: The best practice is to use like this: https://github.com/helm/charts/blob/b2bc5480b16bd29587cbb0a295b35b2306c81dc0/stable/openebs/templates/daemonset-ndm.yaml#L4

PR #1698, opened 2020-05-15
refact(exporter): handle concurrent scrape requests

utkarshmani1997:

This commit fixes an issue where it handles the concurrent http
requests and rejects it if a request is already in progress.

**What this PR does?**:
It adds a new metrics "openebs_target_reject_request_counter" to increase the count
of rejected requests, which can be helpful in debugging and improves mutex locking.

No
- Introduce delay in fake server written in cstor_test.go and send multiple concurrent get requests to
  the exporter, it will only process 1 request at a time and rest other requests will be rejected and new metrics openebs_target_reject_request_counter will be incremented.

No

**Checklist:**
- [x] Fixes https://mdap.zendesk.com/agent/tickets/1073
- [x] Has the change log section been updated? 
- [x] Commit has unit tests

### DIFF
--- a/changelogs/unreleased/1698-utkarshmani1997
+++ b/changelogs/unreleased/1698-utkarshmani1997
@@ -1,1 +1,1 @@
-refact(exporter): handle concurrent GET http requests
+refact(exporter): handle concurrent scrape requests
--- a/changelogs/unreleased/1698-utkarshmani1997
+++ b/changelogs/unreleased/1698-utkarshmani1997
@@ -1,0 +1,1 @@
+refact(exporter): handle concurrent GET http requests
--- a/cmd/maya-exporter/app/collector/collector.go
+++ b/cmd/maya-exporter/app/collector/collector.go
@@ -15,6 +15,8 @@
 package collector
 
 import (
+	"sync"
+
 	v1 "github.com/openebs/maya/pkg/stats/v1alpha1"
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/klog"
@@ -22,8 +24,20 @@ import (
 
 // collector implements prometheus.Collector interface
 type collector struct {
+	sync.Mutex
+	request bool
 	Volume
 	metrics
+}
+
+func (c *collector) isRequestInProgress() bool {
+	return c.request
+}
+
+func (c *collector) setRequestToFalse() {
+	c.Lock()
+	c.request = false
+	c.Unlock()
 }
 
 func New(vol Volume) *collector {
@@ -32,8 +46,8 @@ func New(vol Volume) *collector {
 		klog.Fatal("exiting...")
 	}
 	return &collector{
-		vol,
-		Metrics(typ),
+		Volume:  vol,
+		metrics: Metrics(typ),
 	}
 }
 
@@ -73,6 +87,7 @@ func (c *collector) collectors() []prometheus.Collector {
 		c.healthyReplicaCounter,
 		c.volumeUpTime,
 		c.isClientConnected,
+		c.targetRejectRequestCounter,
 	}
 }
 
@@ -122,9 +137,21 @@ func (c *collector) Collect(ch chan<- prometheus.Metric) {
 		stats       stats
 	)
 
+	c.Lock()
+	if c.isRequestInProgress() {
+		c.targetRejectRequestCounter.Inc()
+		c.targetRejectRequestCounter.Collect(ch)
+		c.Unlock()
+		return
+	}
+
+	c.request = true
+	c.Unlock()
+
 	klog.V(2).Info("Get metrics")
 	metrics := &c.metrics
 	if volumeStats, err = c.get(); err != nil {
+		c.setRequestToFalse()
 		klog.Errorln(err)
 		c.setError(err)
 	}
@@ -138,6 +165,7 @@ func (c *collector) Collect(ch chan<- prometheus.Metric) {
 	for _, col := range c.collectors() {
 		col.Collect(ch)
 	}
+	c.setRequestToFalse()
 }
 
 func (c *collector) setError(err error) {
--- a/cmd/maya-exporter/app/collector/collector.go
+++ b/cmd/maya-exporter/app/collector/collector.go
@@ -25,18 +25,20 @@ import (
 // collector implements prometheus.Collector interface
 type collector struct {
 	sync.Mutex
-	request bool
+	requestInFlight bool
 	Volume
 	metrics
 }
 
-func (c *collector) isRequestInProgress() bool {
-	return c.request
+func (c *collector) isrequestInFlightInProgress() bool {
+	c.Lock()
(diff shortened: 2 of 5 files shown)


cmd/maya-exporter/app/collector/collector.go
```diff
@@ -15,25 +15,39 @@
...
 
 // collector implements prometheus.Collector interface
 type collector struct {
+	sync.Mutex
+	request bool
 	Volume
 	metrics
 }
 
+func (c *collector) isRequestInProgress() bool {
```
mynktl, line 33: Better to add a comment here for caller to take a lock before calling this function.

cmd/maya-exporter/app/collector/collector.go
```diff
@@ -122,9 +137,21 @@ func (c *collector) Collect(ch chan<- prometheus.Metric) {
...
+		return
+	}
+
+	c.request = true
+	c.Unlock()
+
 	klog.V(2).Info("Get metrics")
 	metrics := &c.metrics
 	if volumeStats, err = c.get(); err != nil {
+		c.setRequestToFalse()
```
mynktl, line 154: Do we need to unset request here? If one call errored and second one got succeeded then final metrics may not have expected value in both call. WDYT?

mynktl:
Are we targeting this PR for 1.11?

cmd/maya-exporter/app/collector/collector.go
utkarshmani1997: @mynktl other request will be succeeded only after resetting it, since lock is only protecting the c.request flag. In case of error we want that other blocked requests continue as early as possible.

utkarshmani1997: in other words, until we reset the flags, other requests will be rejected here in this piece of code
```
	c.Lock()
	if c.isRequestInProgress() {
		c.targetRejectRequestCounter.Inc()
		c.targetRejectRequestCounter.Collect(ch)
		c.Unlock()
		return
	}
```

cmd/maya-exporter/app/collector/collector.go
utkarshmani1997: roger that

utkarshmani1997:
> Are we targeting this PR for 1.11?

yes

cmd/maya-exporter/app/collector/collector.go
mynktl: these function will get executed in go routine so we can't be sure of the sequence. Consider following:
routine 1:
c.get() returned error and set request false
routine 2:
executed c.get() since routine 1 set the request to false.

now in both routine are trying to assign stats to c.metrics, one has empty stats and other one has proper. We can't be sure if second routine assigned first or after routine 1.

cmd/maya-exporter/app/collector/cstor_test.go
```diff
@@ -180,6 +183,7 @@ func TestCstorCollector(t *testing.T) {
 			defer resp.Body.Close()
 
 			buf, err := ioutil.ReadAll(resp.Body)
+			fmt.Println(string(buf))
```
kmova, line 186: Is this a debug log?

utkarshmani1997: yes, thanks for pointing it out.

cmd/maya-exporter/app/collector/metrics.go
```diff
@@ -260,6 +261,14 @@ func Metrics(cas string) metrics {
 				Help:      "iSCSI Initiator to jiva target login status: (0, 1): {Not Logged In, Logged In}",
 			},
 		),
+
+		targetRejectRequestCounter: prometheus.NewCounter(
+			prometheus.CounterOpts{
+				Namespace: "openebs",
+				Name:      "target_reject_request_counter",
+				Help:      "Total no of rejected GET http requests if a request is already in progress",
```
kmova, line 269: n_: `Total number of rejected scrape requests if a previous request is in progress`

cmd/maya-exporter/app/collector/collector.go
```diff
@@ -15,25 +15,39 @@
...
+
 	v1 "github.com/openebs/maya/pkg/stats/v1alpha1"
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/klog"
 )
 
 // collector implements prometheus.Collector interface
 type collector struct {
+	sync.Mutex
+	request bool
```
kmova, line 28: n_: request > requestInFlight

cmd/maya-exporter/app/collector/collector.go
kmova: Is it possible to add lock/unlock in this function and avoid the caller from having to worry about the locks?

cmd/maya-exporter/app/collector/collector.go
```diff
@@ -122,6 +137,17 @@ func (c *collector) Collect(ch chan<- prometheus.Metric) {
 		stats       stats
 	)
 
+	c.Lock()
+	if c.isRequestInProgress() {
+		c.targetRejectRequestCounter.Inc()
+		c.targetRejectRequestCounter.Collect(ch)
+		c.Unlock()
```
kmova, line 144: Related to my above comment on pushing lock/unlock to RequestInProgress, will this sequence work?

```
		c.Unlock()
		c.targetRejectRequestCounter.Inc()
		c.targetRejectRequestCounter.Collect(ch)
```

changelogs/unreleased/1698-utkarshmani1997
```diff
@@ -0,0 +1 @@
+refact(exporter): handle concurrent GET http requests
```
kmova, line 1: n_ : handle concurrent scrape requests

cmd/maya-exporter/app/collector/collector.go
utkarshmani1997: channel itself is protected by lock so sequence might change because after unlock we don't know which request came first and which one incremented the the request counter. Though it doesn't matter as we are only incrementing counters, so above sequence will work also.

utkarshmani1997: yes

cmd/maya-exporter/app/collector/collector.go
```diff
@@ -122,6 +139,16 @@ func (c *collector) Collect(ch chan<- prometheus.Metric) {
 		stats       stats
 	)
 
+	if c.isrequestInFlightInProgress() {
+		c.targetRejectRequestCounter.Inc()
+		c.targetRejectRequestCounter.Collect(ch)
+		return
+	}
+
+	c.Lock()
```
kmova, line 148: Can these three lines be in a function -- just like for setting false? 
What is the reason for following different approaches for setting false vs setting to true?

utkarshmani1997: Done

PR #1712, opened 2020-06-08
skip validations checks and handle snapshot deletion once migrated to CSI

prateekpandey14:
## Pull Request template

PR required to handle PVC validation checks before starting the volume migration operations,
for example, ignore the PVC deletion if snapshots are exists.  

**What this PR does?**:
- Add skip validations checks in webhook validations via annotations:
There are some cases like volume migration where we need to delete
the old pvc and create new pvc based on CSI, in such cases we need
to ignore/skip the validations  related to snapshots .

- Ignore old snapshots delete request once migrated to CSI based volumes. Once we migrate the volumes to CSI, we have to migrate the snapshot as well, once snapshots are migrated, backend snapshot will be bind to new CSI snapshot type, and it requires clean up of the older snapshot related resources as they are no longer requried. This change will allow user to delete the snapshot of old types (not from the pools which is bind to the new csi snapshot types)

NO

- Created PVC with skip annotation 
- Created snapshot and tried to delete the pvc which should allow to delete the PVC.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [x] Commit has unit tests
- [x] Commit has integration tests

### DIFF
--- a/pkg/snapshot/v1alpha1/snapshot.go
+++ b/pkg/snapshot/v1alpha1/snapshot.go
@@ -27,6 +27,7 @@ import (
 	"github.com/pkg/errors"
 	v1_storage "k8s.io/api/storage/v1"
 	mach_apis_meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
 )
 
 // options contains the options with respect to
@@ -211,6 +212,17 @@ func (s *snapshot) Delete() (*v1alpha1.CASSnapshot, error) {
 		return nil, err
 	}
 
+	snap := &v1alpha1.CASSnapshot{}
+	// Return success for delete snapshot request in case volume snapshot belongs to a
+	// migrated CSI based persistent volume. Later as part of migration snapshot
+	// can be migrated to CSI snapshots using a migration or manual steps
+	if pv.Spec.CSI != nil {
+		if string(pv.Spec.CSI.Driver) == string(v1alpha1.CSIDriverName) {
+			klog.Infof("skip backend snapshot delete request of migrated CSI volume: %q", s.snapOptions.VolumeName)
+			return snap, nil
+		}
+	}
+
 	storageEngine := pv.Labels[string(v1alpha1.CASTypeKey)]
 	scName := pv.Labels[string(v1alpha1.StorageClassKey)]
 	if len(scName) == 0 {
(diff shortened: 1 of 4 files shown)


pkg/snapshot/v1alpha1/snapshot.go
```diff
@@ -211,6 +211,13 @@ func (s *snapshot) Delete() (*v1alpha1.CASSnapshot, error) {
 		return nil, err
 	}
 
+	// Return success for delete snapshot request in case volume snapshot belongs to a
+	// migrated CSI based persistent volume. Later as part of migration snapshot
+	// can be migrated to CSI snapshots using a migration or manual steps
+	if string(pv.Spec.CSI.Driver) == string(v1alpha1.CSIDriverName) {
```
shubham14bajpai, line 217: This will give nil pointer error for external provisioned volumes with ISCSI spec.

prateekpandey14: ahh missed to add that check, added now

pkg/snapshot/v1alpha1/snapshot.go
```diff
@@ -211,6 +211,15 @@ func (s *snapshot) Delete() (*v1alpha1.CASSnapshot, error) {
 		return nil, err
 	}
 
+	// Return success for delete snapshot request in case volume snapshot belongs to a
+	// migrated CSI based persistent volume. Later as part of migration snapshot
+	// can be migrated to CSI snapshots using a migration or manual steps
+	if pv.Spec.CSI != nil {
+		if string(pv.Spec.CSI.Driver) == string(v1alpha1.CSIDriverName) {
+			return nil, nil
```
mittachaitu, line 219: Did we test this scenario? I have a doubt whether we are returning the response code as 200 or not(If object is nil).

prateekpandey14: We are not checking for nil object, only errors should be nil, so this fine.

mittachaitu: We are checking https://github.com/openebs/maya/blob/c1befaa87a8a708370b9cdc941639d06197a1b9c/cmd/maya-apiserver/app/server/http.go#L450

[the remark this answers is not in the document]
prateekpandey14: added

PR #1808, opened 2021-06-24
feat(analytic): support to use custom server for analytics

mynktl:
**What this PR does?**:
This PR refactors the usage package APIs.

Caller can use the following updated API to send analytics to google:
- To create the usage builder object
         usage.New()
- To build the usage object
        usage.New().Build()
- To send the application analytic
       usageObj.ApplicationBuilder().Send()
- To send the installation analytic
       usageObj.InstallBuilder(true).Send()

**Does this PR require any upgrade changes?**: No

_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests


pkg/usage/send.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/pkg/errors"
+)
+
+// Send the analytic results to server
+// if wait is not set then Send will execute go routine
+func (u *Usage) Send(wait bool) error {
+	var fn func() error
+
+	if u.url == "" {
+		fn = u.SendGoogleAnalytic
```
kmova, line 18: If it is sending to google analytics - let it always be unreliable. attempt to send but don't wait or error on failures. 

In case URL based (reliable service),  then we probably need to do a few things:
- during the initialization, check if a connection can be established and messages sent (this needs to be designed) 
- probably a handshake message needs to be established between the client and server on what parameters needs to be sent
- take additional parameters - to check if sending events should be attempted in the background, handle cases like pod restarts, and have the ability to resend events. 

Give all the above, wondering if we should limit this PR to refactoring changes only? Once we finalize the proposed design for sending events reliable service (or call home), we can add the PRs.

cmd/maya-apiserver/app/server/volume_endpoint_v1alpha1.go
```diff
@@ -57,14 +58,20 @@ type volumeAPIOpsV1alpha1 struct {
...
 			SetLabel(usage.EventLabelCapacity).
 			SetReplicaCount(cvol.Spec.Replicas, method).
 			SetCategory(method).
-			SetVolumeCapacity(cvol.Spec.Capacity).Send()
+			SetVolumeCapacity(cvol.Spec.Capacity).
+			Build()
+		if err != nil {
+			klog.Errorf("Error fetching analytic object err=%s", err)
+		}
+		obj.ApplicationBuilder().Send()
```
vharsh, line 74: should this be run in a separate go-routine like earlier?

pkg/usage/usage.go
```diff
@@ -118,6 +122,7 @@ func (u *Usage) SetTrackingID(track string) *Usage {
 
 // SetCampaignName : set the name of the PVC or will be empty.
 func (u *Usage) SetCampaignName(campaignName string) *Usage {
+	fmt.Printf("setting asoadoas %s\n", campaignName)
```
vharsh, line 125: You might have missed removing this debug log.

```suggestion
```

pkg/usage/usage.go
```diff
@@ -149,8 +154,13 @@ func (u *Usage) SetApplicationID(appID string) *Usage {
...
 // SetApplicationVersion : usecase(project-version)
-func (u *Usage) SetApplicationVersion(appVersion string) *Usage {
-	u.appVersion = appVersion
+func (u *Usage) SetApplicationVersion(appVersion, gitcommit string) *Usage {
+	if gitcommit == "" {
+		u.appVersion = appVersion
+		return u
+	}
+
+	u.appVersion = strings.Join([]string{appVersion, gitcommit[0:7]}, "-")
```
vharsh, line 163: (optional) This might look simpler, if strings.Join does not offer any kind of nil safety operations,  considering appVersion cannot be nil it has to be `""` at worst.

```suggestion
	u.appVersion = appVersion + "-" + gitcommit[0:7]
```

vharsh (commented):
minor optional suggestions & nitpicks

pkg/usage/usage.go
```diff
@@ -250,17 +234,45 @@ func (u *Usage) SetReplicaCount(count, method string) *Usage {
...
 }
 
+// Build is a builder method for Usage struct
+func (u *Usage) Build() (*Usage, error) {
+	// Default ApplicationID for openebs project is OpenEBS
+	v, err := NewVersion()
+	if err != nil {
+		return nil, err
+	}
+
```
vharsh, line 244: has k8s version been set somewhere else?

pkg/usage/send.go
vharsh: Looks like this file has been removed from file changes @ `4b00533: Removing sending analytic to REST endpoint`, 
I couldn't find it under the `Files changed` tab, if this has been moved to upcoming PRs, please also update the PR heading, just as the PR description has been updated.

pkg/usage/googleanalytics.go
```diff
@@ -17,37 +17,31 @@ package usage
...
+		ApplicationVersion(u.appVersion).
+		DataSource(u.dataSource).
+		ApplicationName(u.appName).
+		ApplicationInstallerID(u.appInstallerID).
+		DocumentTitle(u.documentTitle)
+	// Un-wrap the Event struct back here
+	event := analytics.NewEvent(u.category, u.action)
+	event.Label(u.label)
+	event.Value(u.value)
+	return gaClient.Send(event)
```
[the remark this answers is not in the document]
mynktl: >Is it that this function will be called in a new go-routine if required by the caller

yes

pkg/usage/usage.go
mynktl: yes, we can convert it since there are two args.

pkg/usage/usage.go
```diff
@@ -250,17 +234,45 @@ func (u *Usage) SetReplicaCount(count, method string) *Usage {
...
+	v.setOpenEBSVersion(u.appVersion)
+	v.getVersion(false)
+	u.version = v
+	u.SetApplicationID(AppName).
+		SetTrackingID(GAclientID).
+		SetClientID(v.id).
+		SetCampaignSource(v.installerType)
+	// TODO: Add condition for version over-ride
+	// Case: CAS/Jiva version, etc
```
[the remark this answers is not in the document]
mynktl: it is being set in `getVersion` function.

vharsh (approved):
lgtm

kmova (approved):

vharsh:
> This PR needs update in all the dependent repos. This will be taken up as part of the next release.

Is moving it out into a separate feature-repo not in the plan, such that this becomes a vendor-ed library so that effort is reduced to a simple dependency version bump up?
Was the earlier plan modified because of different ways each module will expose the `.version` variable location in constants of the binary?

kmova:
this is not required anymore.

PR #1813, opened 2021-08-21
fix(crd): migration from v1beta1 to v1

prateekpandey14:
Change the version in the CRDs to v1 and make the
necessary movements of the elements to pass the
CRD v1 validations like:
- rename version to versions
- set the defaults for a specific version
- set the schema validation to `x-kubernetes-preserve-unknown-fields`

Also, this PR removes the installation of CRDs that are already
migrated to cstor CSI:
- cstorpoolinstances
- cstorvolumeclaims
- cstorvolumepolicies
cherrypick:#1812

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ jobs:
     - os: linux
       arch: amd64
       env:
-        - RELEASE_TAG_DOWNSTREAM=1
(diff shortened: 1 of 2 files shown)


niladrih (approved):
lgtm

PR #1823, opened 2021-09-14
feat(cas-template-engine): add 'data' map key to volumeconfig alongside value and enabled

niladrih:
Signed-off-by: Niladri Halder <niladri.halder@mayadata.io>

This is required in openebs/dynamic-localpv-provisioner repo to parse the cas-template of format
```yaml
- name: XFSQuota
  enabled: true
  data: |
      SoftLimitGrace: 80
      HardLimitGrace: 85
```

**What this PR does?**:
Adds instruction to add the embedded Data component from v1alpha1.config object to the map when ConfigToMap function is called.


niladrih:
This change breaks existing code. Closing this.

